PR #20, opened 2024-10-10
Update README and Contribution guide

zRzRzRzRzRzRzR:


README.md
```diff
@@ -2,7 +2,7 @@
 
 [中文阅读](./README_zh.md)
 
-Fine-tune Cog family of video models for custom video generation under 24GB of GPU memory ⚡️📼
+Fine-tune Cog family of video models for custom video generation under 24GB of GPU memory, support multi resolutions ⚡️📼
```
a-r-r-o-w, line 5: I think for headline we can just leave it to what it was before. There can be a separate section for supported capabilities once we add a few more things

```suggestion
Fine-tune Cog family of video models for custom video generation under 24GB of GPU memory ⚡️📼
```

README.md
```diff
@@ -23,7 +23,8 @@ huggingface-cli download \
   --local-dir video-dataset-disney
 ```
 
-Then launch LoRA fine-tuning for text-to-video (modify the different hyperparameters, dataset root, and other configuration options as per your choice):
+Then launch LoRA fine-tuning for text-to-video (modify the different hyperparameters, dataset root, and other
```
a-r-r-o-w, line 26: Is there a reason why we've limited the line length? It makes it a bit harder to review all the changes and seems unnecessary because line wrapping doesn't really affect the GitHub README view

README.md
```diff
@@ -53,31 +55,56 @@ video = pipe("<my-awesome-prompt>").frames[0]
...
 
 ## Dataset Preparation
 
-Create two files where one file contains line-separated prompts and another file contains line-separated paths to video data (the path to video files must be relative to the path you pass when specifying `--data_root`). Let's take a look at an example to understand this better!
+Create two files: one containing line-separated prompts, and the other containing line-separated paths to video files (
+video file paths must be relative to the path you specify with `--data_root`). Let's understand this better with an
+example!
 
-Assume you've specified `--data_root` as `/dataset`, and that this directory contains the files: `prompt.txt` and `videos.txt`.
+Assume your specified `--data_root` is `/dataset`, and this directory contains the files `prompts.txt` and `videos.txt`.
```
a-r-r-o-w, line 71: I think better to keep as `prompt.txt` for the moment since the example dataset we're using (Disney) contains the filename as `prompt.txt`. If we can change the filename on the HF dataset repository, we can update it here as well to `prompts.txt`

README.md
```diff
@@ -53,31 +55,56 @@ video = pipe("<my-awesome-prompt>").frames[0]
...
-The `videos.txt` file should contain line-separate paths to video files. Note that the path should be _relative_ to the `--data_root` directory.
+### Preparing the Video Dataset
+
+The framework supports the following specific resolutions and frame counts:
+
+- **Resolutions (Width * Height) Supported**:
+    - Any width * height divisible by 32 between 256 and 2048. For example, `720 * 480`, `1920 * 1020`.
+
+- **Frame Counts (Frames) Supported**:
+    - 16, 24, 32, 48, 64, 80. The frame count must be divisible by 4.
```
a-r-r-o-w, line 95: I think better to say `4 * k` or `4 * k + 1`

README.md
```diff
@@ -53,31 +55,56 @@ video = pipe("<my-awesome-prompt>").frames[0]
...
+
 ```
 
-The `videos.txt` file should contain line-separate paths to video files. Note that the path should be _relative_ to the `--data_root` directory.
+### Preparing the Video Dataset
+
+The framework supports the following specific resolutions and frame counts:
+
+- **Resolutions (Width * Height) Supported**:
+    - Any width * height divisible by 32 between 256 and 2048. For example, `720 * 480`, `1920 * 1020`.
```
a-r-r-o-w, line 92: This is actually incorrect.

For Text-to-video:
- In 2B model, we use static positional embeddings. So the model needs to initialized with different `sample_height`, `sample_width` and `sample_frames`.
- In 5B model, we use RoPE. So the model can use any resolution pretty much with base resolution as `720 x 480`.

For Image-to-video:
- We only have 5B model. It uses fixed learned positional embeddings so it is quite limited in resolution size. I have noticed that the learned positional embedding values are close to 0 (mean is very close to 0 and variance is extremely small), and even if we remove that layer, generation works quite well. In the future, we could add an option to remove that layer while training and just use RoPE (which can scale to all kinds of resolutions)

README.md
zRzRzRzRzRzRzR: Oh, this is because I used PyCharm for editing with its automatic formatting (the display for lines exceeding the length might have gone beyond my screen, and I think some developers have the same issue, so I used the IDE's auto-formatting, which might need adjustment

README.md
zRzRzRzRzRzRzR: use your's is correct

README.md
zRzRzRzRzRzRzR: But I observed that there are actually more restrictions in this part of the code
```
HEIGHT_BUCKETS = [256, 320, 384, 480, 512, 576, 720, 768, 960, 1024, 1280, 1536]
WIDTH_BUCKETS = [256, 320, 384, 480, 512, 576, 720, 768, 960, 1024, 1280, 1536]
FRAME_BUCKETS = [16, 24, 32, 48, 64, 80]
```
In this implementation, the resolution is limited by the three parameters HEIGHT_BUCKETS, WIDTH_BUCKETS, and FRAME_BUCKETS. When processing video, the system will categorize the video's frame count, width, and height into the nearest bucket based on these buckets. If the video's resolution or frame count is not within these buckets, the code will find the bucket closest to the video.
Therefore, if it exceeds 1536, it should be adjusted to 1536; a 1920x1080 video will be adjusted to 1536x1024 for processing. This means the video's height will be adjusted to 1024 and the width to 1536. This is my understanding; should the readme be modified in this way?

README.md
sayakpaul: Agree with the suggestion.

README.md
sayakpaul: Yeah `prompt.txt` should be fine for now. This way we don't have to update the remote artifact (dataset in our case).

README.md
```diff
@@ -53,31 +55,56 @@ video = pipe("<my-awesome-prompt>").frames[0]
...
-Assume you've specified `--data_root` as `/dataset`, and that this directory contains the files: `prompt.txt` and `videos.txt`.
+Assume your specified `--data_root` is `/dataset`, and this directory contains the files `prompts.txt` and `videos.txt`.
 
-The `prompt.txt` file should contain line-separated prompts:
+### Preparing the Prompts Dataset
+
+The `prompts.txt` file should contain line-separated prompts. Please note that the prompts must be in English, and it is
+recommended to use the [prompt polishing script](https://github.com/THUDM/CogVideo/blob/main/inference/convert_demo.py)
+for refinement. Alternatively, you can use [CogVideo-caption](https://huggingface.co/THUDM/cogvlm2-llama3-caption) for
+captioning:
```
sayakpaul, line 78: Provided we're sticking to `prompt.txt`:

```suggestion
The `prompt.txt` file should contain line-separated prompts. Please note that the prompts must be in English, and it is
recommended to use the [prompt refinement script](https://github.com/THUDM/CogVideo/blob/main/inference/convert_demo.py)
for better prompts. Alternatively, you can use [CogVideo-caption](https://huggingface.co/THUDM/cogvlm2-llama3-caption) for
captioning:
```

README.md
sayakpaul: We should then also denote what is "k". I think the current proposed change is good enough.

README.md
sayakpaul: We could maybe move this to "dataset.md" guide and reference it from the main README?

README.md
```diff
@@ -128,7 +170,10 @@ We provide training script for both text-to-video and image-to-video generation
   MAX_TRAIN_STEPS=("3000")
   ```
 
-- Select which Accelerate configuration you would like to train with: `ACCELERATE_CONFIG_FILE="accelerate_configs/uncompiled_1.yaml"`. We provide some default configurations in the `accelerate_configs/` directory - single GPU uncompiled/compiled, 2x GPU DDP, DeepSpeed, etc. You can create your own config files with custom settings using `accelerate config --config_file my_config.yaml`.
+- Select which Accelerate configuration you would like to train
+  with: `ACCELERATE_CONFIG_FILE="accelerate_configs/uncompiled_1.yaml"`. We provide some default configurations in
+  the `accelerate_configs/` directory - single GPU uncompiled/compiled, 2x GPU DDP, DeepSpeed, etc. You can create your
+  own config files with custom settings using `accelerate config --config_file my_config.yaml`.
```
sayakpaul, line 176: Is this because of your PyCharm formatting?

README.md
```diff
@@ -196,9 +241,11 @@ We provide training script for both text-to-video and image-to-video generation
   done
   ```
 
-  To understand what the different parameters mean, you could either take a look at the [args](./training/args.py) file or run the training script with `--help`.
+  To understand what the different parameters mean, you could either take a look at the [args](./training/args.py) file
+  or run the training script with `--help`.
```
sayakpaul, line 245: Is this some indentation change?

README.md
```diff
@@ -196,9 +241,11 @@ We provide training script for both text-to-video and image-to-video generation
...
 
-Note: Training scripts are untested on MPS, so performance and memory requirements can differ widely compared to the CUDA reports below.
+Note: Training scripts are untested on MPS, so performance and memory requirements can differ widely compared to the
+CUDA reports below.
```
sayakpaul, line 248:
```suggestion
> [!NOTE]  
> Training scripts are untested on MPS, so performance and memory requirements can differ widely compared to the CUDA reports below.
```

README.md
```diff
@@ -230,70 +277,89 @@ Note: Training scripts are untested on MPS, so performance and memory requiremen
...
 
 > [!IMPORTANT]
-> The memory requirements are reported after running the `training/prepare_dataset.py`, which converts the videos and captions to latents and embeddings. During training, we directly load the latents and embeddings, and do not require the VAE or the T5 text encoder. However, if you perform validation/testing, these must be loaded and increase the amount of required memory. Not performing validation/testing saves a significant amount of memory, which can be used to focus solely on training if you're on smaller VRAM GPUs.
+> The memory requirements are reported after running the `training/prepare_dataset.py`, which converts the videos and
+> captions to latents and embeddings. During training, we directly load the latents and embeddings, and do not require
+> the
+> VAE or the T5 text encoder. However, if you perform validation/testing, these must be loaded and increase the amount
+> of
+> required memory. Not performing validation/testing saves a significant amount of memory, which can be used to focus
+> solely on training if you're on smaller VRAM GPUs.
```
sayakpaul, line 299: Why do we need this?

README.md
```diff
@@ -230,70 +277,89 @@ Note: Training scripts are untested on MPS, so performance and memory requiremen
...
+> script, or extract some frames from your training data using:
 > `ffmpeg -i input.mp4 -frames:v 1 frame.png`,
 > or provide a URL to a valid and accessible image.
 
 <details>
 <summary> AdamW </summary>
 
-**Note:** Trying to run CogVideoX-5b without gradient checkpointing OOMs even on an A100 (80 GB), so the memory measurements have not been specified.
+**Note:** Trying to run CogVideoX-5b without gradient checkpointing OOMs even on an A100 (80 GB), so the memory
+measurements have not been specified.
```
sayakpaul, line 319: No change here no?

README.md
```diff
@@ -230,70 +277,89 @@ Note: Training scripts are untested on MPS, so performance and memory requiremen
...
 ### LoRA finetuning
 
 > [!NOTE]
-> The memory requirements for image-to-video lora finetuning are similar to that of text-to-video on `THUDM/CogVideoX-5b`, so it hasn't been reported explicitly.
+> The memory requirements for image-to-video lora finetuning are similar to that of text-to-video
+> on `THUDM/CogVideoX-5b`, so it hasn't been reported explicitly.
 >
-> Additionally, to prepare test images for I2V finetuning, you could either generate them on-the-fly by modifying the script, or extract some frames from your training data using:
+> Additionally, to prepare test images for I2V finetuning, you could either generate them on-the-fly by modifying the
+> script, or extract some frames from your training data using:
```
sayakpaul, line 311: Same. Why do we need this?

README.md
```diff
@@ -230,70 +277,89 @@ Note: Training scripts are untested on MPS, so performance and memory requiremen
...
 
-**Note:** Trying to run CogVideoX-5b without gradient checkpointing OOMs even on an A100 (80 GB), so the memory measurements have not been specified.
+**Note:** Trying to run CogVideoX-5b without gradient checkpointing OOMs even on an A100 (80 GB), so the memory
+measurements have not been specified.
 
 With `train_batch_size = 1`:
 
 |       model        | lora rank | gradient_checkpointing | memory_before_training | memory_before_validation | memory_after_validation | memory_after_testing |
 |:------------------:|:---------:|:----------------------:|:----------------------:|:------------------------:|:-----------------------:|:--------------------:|
-| THUDM/CogVideoX-2b |    16     |          False         |         12.945         |          43.764          |         46.918          |       24.234         |
```
sayakpaul, line 260: Could leave this unchanged I guess?

README.md
```diff
@@ -480,9 +554,12 @@ With `train_batch_size = 4`:
...
-- [ ] Test scripts with AdamW (8-bit bitsandbytes) + CPUOffloadOptimizer (with gradient offloading) (Currently errors out)
-- [ ] [Sage Attention](https://github.com/thu-ml/SageAttention) (work with the authors to support backward pass, and optimize for A100)
+- [ ] Test scripts with torchao quantization, and low bit memory optimizers (Currently errors with AdamW (8/4-bit
+  torchao))
+- [ ] Test scripts with AdamW (8-bit bitsandbytes) + CPUOffloadOptimizer (with gradient offloading) (Currently errors
+  out)
+- [ ] [Sage Attention](https://github.com/thu-ml/SageAttention) (work with the authors to support backward pass, and
+  optimize for A100)
 
 > [!IMPORTANT]
```
sayakpaul, line 564: Let's include note about our contribution guide?

sayakpaul (commented):
Thanks for your PR! I like the changes. I think we can keep the formatting related changes to a bare minimum but the informative content -- let's make sure to include them.

README.md
a-r-r-o-w: > But I observed that there are actually more restrictions in this part of the code

Those are just the default values if any are not specified by the user actually. We support three different parameters that users should be able to customize: `--height_buckets`, `--width_buckets`, `--frame_buckets`.

So, let's say I wanted to train at 81 x 2048 x 1024 and 49 x 512 x 768 videos, I could run the training with: `--height_buckets 1024 768 --width_buckets 2048 512 --frame_buckets 81 x 49`.

I think if we could better explain this, it would be helpful. But, as such, we are not limited by the frame resolutions that have been hardcoded there

README.md
a-r-r-o-w: @zRzRzRzRzRzRzR Could you revert the PyCharm introduced formatting, please? I think there should also be a setting to disable it. It would be nice to not have line wrapping within the README itself because GitHub view automatically handles this

README.md
zRzRzRzRzRzRzR: > We could maybe move this to "dataset.md" guide and reference it from the main README?我们也许可以把这部分移动到"dataset.md"指南中，并从主 README 中引用它？

I think it is a better idea

zRzRzRzRzRzRzR:
I have reverted it back to its original form and created a new file called "dataset.md".

README.md
```diff
@@ -57,54 +57,9 @@ export_to_video(video, "output.mp4", fps=8)
...
-
-## Training
+## Prepare Dataset and Training
 
+Before starting the training, please check whether the dataset has been prepared according to the [dataset specifications](assets/dataset.md). We provide training scripts suitable for text-to-video and image-to-video generation, compatible with the [CogVideoX model family](https://huggingface.co/collections/THUDM/cogvideo-66c08e62f1685a3ade464cce). Training can be started using the `train*.sh` scripts, depending on the task you want to train. Let's take LoRA fine-tuning for text-to-video as an example.
 We provide training script for both text-to-video and image-to-video generation which are compatible with the [CogVideoX family of models](https://huggingface.co/collections/THUDM/cogvideo-66c08e62f1685a3ade464cce). Training can be launched with one of the `train*.sh` scripts based on the task you'd like to train. Let's take text-to-video LoRA finetuning as an example.
```
a-r-r-o-w, line 63:
```suggestion
```

a-r-r-o-w (approved):
thanks!

PR #24, opened 2024-10-11
Update requirements.txt (fixed typo)

Nojahhh:
Fixed typo transformers dependancy from 0.45.2 to 4.45.2


a-r-r-o-w (approved):
Thank you! I think the requirements version were pushed in a rush, so might be incorrect. I'll take a look and fix everything after this PR

PR #27, opened 2024-10-11
Lower requirements versions

a-r-r-o-w:
Lowers the requirements a bit to make it more friendly with existing user environments


zRzRzRzRzRzRzR (commented):
"I believe there are significant changes in numpy 2.X. Should we pin the version to 1.26.4, which is the last version of numpy 1.x? However, I am not sure if there will be compatibility issues between these two versions."

requirements.txt
```diff
@@ -1,16 +1,16 @@
-accelerate>=1.0.0
-bitsandbytes>=0.44.1
-diffusers>=0.30.4
+accelerate>=0.34.2
+bitsandbytes>=0.43.3
```
sayakpaul, line 2: I think we can definitely unpin the version. It's safe.

requirements.txt
```diff
@@ -1,16 +1,16 @@
...
+diffusers>=0.30.3
 transformers>=4.45.2
-huggingface_hub>=0.25.2
+huggingface_hub>=0.24.6
```
sayakpaul, line 5: Same here.

requirements.txt
```diff
@@ -1,16 +1,16 @@
...
-pandas>=2.4.1
+wandb>=0.17.7
+pandas>=2.2.2
 torch>=2.4.0
 torchvision>=0.19.0
 torchao>=0.5.0
 sentencepiece>=0.2.0
 imageio-ffmpeg>=0.5.1
-numpy>=2.1.2
+numpy>=1.26.4
```
sayakpaul, line 16: Good one.

requirements.txt
```diff
@@ -1,16 +1,16 @@
...
-huggingface_hub>=0.25.2
+huggingface_hub>=0.24.6
 hf_transfer>=0.1.8
-peft>=0.13.1
+peft>=0.12.0
 decord>=0.6.0
-wandb>=0.18.3
-pandas>=2.4.1
+wandb>=0.17.7
+pandas>=2.2.2
```
sayakpaul, line 10: Same here.

sayakpaul (approved):
Thanks! Just a single comment.

a-r-r-o-w:
> I believe there are significant changes in numpy 2.X. Should we pin the version to 1.26.4, which is the last version of numpy 1.x? However, I am not sure if there will be compatibility issues between these two versions.

I have tried experimenting with both versions so far and not found any problems. I think we use very basic functionalities, and do most of things in torch land, so we should be good hopefully with this 1.26.4 version. Please let me know if you find any problems if you'd like to try

requirements.txt
a-r-r-o-w: Sounds good!

a-r-r-o-w, line 2:
```suggestion
accelerate
bitsandbytes
```

requirements.txt
a-r-r-o-w, line 5:
```suggestion
huggingface_hub
```

requirements.txt
a-r-r-o-w, line 10:
```suggestion
wandb
pandas
```

a-r-r-o-w:
Made some updates based on reviews, please approve if okay and merge

PR #32, opened 2024-10-15
Update for windows compability

Nojahhh:

PR #48, opened 2024-10-18
Windows support for T2V scripts

a-r-r-o-w:
@Nojahhh Would you mind giving this a review? You've already fixed the dataset side of things with lambda pickling error, and this PR replaces the inline collate_fn with a class instance as per your PR #32 in the remaining scripts

### DIFF
--- a/training/cogvideox_text_to_video_lora.py
+++ b/training/cogvideox_text_to_video_lora.py
@@ -25,6 +25,7 @@ from typing import Any, Dict
 import diffusers
 import torch
 import transformers
+import wandb
 from accelerate import Accelerator, DistributedType
 from accelerate.logging import get_logger
 from accelerate.utils import (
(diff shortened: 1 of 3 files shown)


training/cogvideox_text_to_video_lora.py
```diff
@@ -199,6 +198,26 @@ def log_validation(
     return videos
 
 
+class CollateFunction:
```
sayakpaul, line 201: Mind explaining the change?

sayakpaul (approved):
Looks minimal .

Nojahhh:
@a-r-r-o-w I think it looks good. Will try it out and report back if I encounter any errors.

training/cogvideox_text_to_video_lora.py
Nojahhh: Inline function raised an error with pickle in windows, moving collate function to a class instances solves the issue.

Nojahhh (approved):
All three training methods have been tested and work as intended.

a-r-r-o-w:
Thanks for all your help @Nojahhh!

PR #51, opened 2024-10-19
fix: correct typo in GitHub files

DhanushNehru:


sayakpaul (commented):
None of these are typos.

PR #54, opened 2024-10-19
Fix gradient accumulation related unbound errors for tracking gradnorm

### LINKED ISSUE #41: cannot access local variable 'gradient_norm_before_clip' where it is not associated with a value
During both I2V and t2V training, sometimes I encountered the error 

```
[rank1]:   File "/root/projects/cogvideox-factory/training/cogvideox_text_to_video_lora.py", line 762, in main
[rank1]:     "gradient_norm_before_clip": gradient_norm_before_clip,
[rank1]:                                  ^^^^^^^^^^^^^^^^^^^^^^^^^
[rank1]: UnboundLocalError: cannot access local variable 'gradient_norm_before_clip' where it is not associated with a value
```

This is probably [here](https://github.com/a-r-r-o-w/cogvideox-factory/blob/a6c246c29d11d78e4aa3fb4b137c5ffd8d719d94/training/cogvideox_text_to_video_lora.py#L715) in the following code
```
if accelerator.sync_gradients:
        gradient_norm_before_clip = get_gradient_norm(transformer.parameters())
        accelerator.clip_grad_norm_(transformer.parameters(), args.max_grad_norm)
        gradient_norm_after_clip = get_gradient_norm(transformer.parameters())
```
somehow `accelerator.sync_gradients` is false sometimes. 

Is there a quick fix? Is it only for logging?

a-r-r-o-w:
Fixes #41. @Yuancheng-Xu would you mind giving this a review?

### DIFF
--- a/training/cogvideox_image_to_video_lora.py
+++ b/training/cogvideox_image_to_video_lora.py
@@ -651,6 +651,8 @@ def main(args):
 
         for step, batch in enumerate(train_dataloader):
             models_to_accumulate = [transformer]
+            gradient_norm_before_clip = None
+            gradient_norm_after_clip = None
 
             with accelerator.accumulate(models_to_accumulate):
                 images = batch["images"].to(accelerator.device, non_blocking=True)
@@ -805,7 +807,7 @@ def main(args):
             last_lr = lr_scheduler.get_last_lr()[0] if lr_scheduler is not None else args.learning_rate
             logs = {"loss": loss.detach().item(), "lr": last_lr}
             # gradnorm + deepspeed: https://github.com/microsoft/DeepSpeed/issues/4555
-            if accelerator.distributed_type != DistributedType.DEEPSPEED:
+            if accelerator.distributed_type != DistributedType.DEEPSPEED and gradient_norm_before_clip is not None:
                 logs.update(
                     {
                         "gradient_norm_before_clip": gradient_norm_before_clip,
--- a/training/cogvideox_text_to_video_lora.py
+++ b/training/cogvideox_text_to_video_lora.py
@@ -635,6 +635,8 @@ def main(args):
 
         for step, batch in enumerate(train_dataloader):
             models_to_accumulate = [transformer]
+            gradient_norm_before_clip = None
+            gradient_norm_after_clip = None
 
             with accelerator.accumulate(models_to_accumulate):
                 videos = batch["videos"].to(accelerator.device, non_blocking=True)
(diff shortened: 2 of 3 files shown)


training/cogvideox_image_to_video_lora.py
```diff
@@ -651,6 +651,8 @@ def load_model_hook(models, input_dir):
 
         for step, batch in enumerate(train_dataloader):
             models_to_accumulate = [transformer]
+            gradient_norm_before_clip = None
```
sayakpaul, line 654: Oh I thought we always calculate it?
https://github.com/a-r-r-o-w/cogvideox-factory/blob/6c00cf094b2f124a61775276d934e483b53b53fb/training/cogvideox_image_to_video_lora.py#L763

sayakpaul (commented):
Left a small comment.

Yuancheng-Xu:
The current fix looks fine to me.

However I have a question: Why not just put all the logging inside `if accelerator.sync_gradients:`?

What I mean is to change the original code

```
if accelerator.sync_gradients:
   progress_bar.update(1)
   global_step += 1
   <some code...>
   
last_lr = lr_scheduler.get_last_lr()[0] if lr_scheduler is not None else args.learning_rate
logs = {
                "loss": loss.detach().item(),
                "lr": last_lr,
                "gradient_norm_before_clip": gradient_norm_before_clip,
                "gradient_norm_after_clip": gradient_norm_after_clip,
            }
progress_bar.set_postfix(**logs)
accelerator.log(logs, step=global_step)
```
to this modified one
```
if accelerator.sync_gradients:
   progress_bar.update(1)
   global_step += 1
   <some code...>

   last_lr = lr_scheduler.get_last_lr()[0] if lr_scheduler is not None else args.learning_rate
   logs = {
                "loss": loss.detach().item(),
                "lr": last_lr,
                "gradient_norm_before_clip": gradient_norm_before_clip,
                "gradient_norm_after_clip": gradient_norm_after_clip,
            }
    progress_bar.set_postfix(**logs)
    accelerator.log(logs, step=global_step)
```

It seems that since the logging only happens when `accelerator.log(logs, step=global_step)`, it makes more sense to only log when `global_step+=1`.

I am not so sure about this, since I also saw the original implementation in other diffuser implementation too, like [here](https://github.com/huggingface/diffusers/blob/2a1d2f62180aa01637b4c1d98e36c7570fd06b1f/examples/dreambooth/train_dreambooth.py#L1382)

sayakpaul:
Not a bad idea but I guess it's more a structural preference and logical convenience honestly.

sayakpaul:
Let's resolve the conflict and merge this @a-r-r-o-w?

a-r-r-o-w:
Yes, sorry about the delay here. I will take a look over the weekend and fix some of the minor bug reports we've received

PR #61, opened 2024-10-20
Fixed optimizers parsing error in bash scripts

Nojahhh:
Fixed an error regarding the first value in optimizers param since bash doesn't use comma to separate multiple values in a list.

The result before was that the script would tell you that there is no such optimizer named "adamw,".


a-r-r-o-w (approved):
thanks!

PR #66, opened 2024-10-22
add some script of lora test

zRzRzRzRzRzRzR:


tests/test_lora_inference.py
```diff
@@ -0,0 +1,42 @@
...
+"""
+
+import argparse
+import torch
+from diffusers import CogVideoXPipeline
+from diffusers.utils import export_to_video
+
+
+def generate_video(prompt, lora_path, lora_name, output_file, fps):
+    pipe = CogVideoXPipeline.from_pretrained("THUDM/CogVideoX-5b", torch_dtype=torch.bfloat16).to("cuda")
```
sayakpaul, line 17: I think this should be supplied too, no? What if someone used the 2B variant?

tests/test_lora_inference.py
```diff
@@ -0,0 +1,42 @@
+"""
+Run this test in Lora adpater checking:
+
+```shell
+python3 test_lora_inference.py --prompt "Your prompt" --lora_path "path/to/lora" --lora_name "lora_adapter" --output_file "output.mp4" --fps 8
```
sayakpaul, line 5: Maybe we use a working command as an example here? 

Path to LoRA can either be local or a Hub id.

sayakpaul (approved):
Neat!

tests/test_lora_inference.py
zRzRzRzRzRzRzR: "Yes, I can turn the model ID into a variable and then add it in."

sayakpaul (approved):
LGTM! @a-r-r-o-w WDYT?

requirements.txt
```diff
@@ -8,8 +8,8 @@ peft>=0.12.0
 decord>=0.6.0
 wandb
 pandas
-torch>=2.4.0
-torchvision>=0.19.0
+torch>=2.5.0
```
a-r-r-o-w, line 11: I think this will not work at the moment because of the cuDNN errors: https://github.com/huggingface/diffusers/issues/9704

```suggestion
torch<2.5.0
```

requirements.txt
```diff
@@ -8,8 +8,8 @@ peft>=0.12.0
...
+torchvision>=0.20.0
```
a-r-r-o-w, line 12:
```suggestion
torchvision<0.20.0
```

Same reason as above

a-r-r-o-w (approved):
Looks good, but I think versions will create a problem currently until pytorch releases 2.5.1

PR #84, opened 2024-11-12
sft with multigpu

zhipuch:
sft with multigpu and gradient accumulation, gradient_norm_before_clip undefined error


sayakpaul (approved):
Thank you!

glide-the:
It seems that there is a bug in the accelerator.is_main_process task and accelerator.distributed_type. Is_main_process has scheduling problems during the training initiation and training stages.
Line 540 also has such a problem.
tracker_name = args.tracker_name or "cogvideox-sft"

PR #34, opened 2024-10-15
captioning + dataset preparation + inference + improvements

sayakpaul:
Known gotchas:

* Not all models listed in https://docs.vllm.ai/en/latest/models/supported_models.html will have the same format for doing video captioning. Change the `recaption.py` as needed to suit your needs. 
* `limit_mm_per_prompt` needs adjustment based on the model being selected. 

After I ran `launch.sh`, I got:

```bash
ls -lh $output_dir/*.txt

...
-rw-rw-r-- 1 sayak sayak  596 oct.  15 11:42 video-dataset-disney/d3af95bf8f9cce29e5c99e839e630c59_caption.txt
-rw-rw-r-- 1 sayak sayak  596 oct.  15 11:42 video-dataset-disney/d8a33062ef6a446c168acb06bf45b77d_caption.txt
-rw-rw-r-- 1 sayak sayak   55 oct.  15 11:42 video-dataset-disney/ec80a7b9cfb3740e5cdf9300fdc8d8a9_caption.txt
-rw-rw-r-- 1 sayak sayak  647 oct.  15 11:42 video-dataset-disney/f6a95cd4397b6a102e82becdcd05d585_caption.txt
-rw-rw-r-- 1 sayak sayak  591 oct.  15 11:42 video-dataset-disney/fe0efaedf8c47812bff8da9951f77975_caption.txt
-rw-rw-r-- 1 sayak sayak  414 oct.  15 11:42 video-dataset-disney/fec5a5184b05acafc7904cd419cbb5a3_caption.txt
```

PR #90, opened 2024-11-17
[feat] add Mochi-1 trainer

sayakpaul:
A minimal and a simple reimplementation of the Mochi-1 fine-tuner but with `diffusers` and `peft`. 

Follow the `README.md` file added in this PR. 

Successful runs will be at https://wandb.ai/sayakpaul/mochi-1-lora.


training/mochi-1/args.py
```diff
@@ -0,0 +1,474 @@
+import argparse
```
a-r-r-o-w, line 1: Okay to have a separate file for now for faster iterations. We will be refactoring the repo with a more modular API in the future anyway

training/mochi-1/dataset.py
```diff
@@ -0,0 +1,444 @@
...
+
+# Must import after torch because this can sometimes lead to a nasty segmentation fault, or stack smashing error
+# Very few bug reports but it happens. Look in decord Github issues for more relevant information.
+import decord  # isort:skip
+
+decord.bridge.set_bridge("torch")
+
+logger = get_logger(__name__)
+
+# TODO (sayakpaul): probably not all buckets are needed for Mochi-1?
```
a-r-r-o-w, line 24: These are just default buckets if you don't specify via CLI args, so I don't think we need to worry here

training/mochi-1/dataset.py
```diff
@@ -0,0 +1,444 @@
...
+            return self._load_preprocessed_latents_and_embeds(path)
+        else:
+            video_reader = decord.VideoReader(uri=path.as_posix())
+            video_num_frames = len(video_reader)
+
+            nearest_frame_bucket = min(
+                self.frame_buckets, key=lambda x: abs(x - min(video_num_frames, self.max_num_frames))
+            )
+            if video_num_frames < nearest_frame_bucket:
+                # TODO: we could handle this by padding zero frames or duplicating the existing frames?
```
a-r-r-o-w, line 294: This will actually break our "belief" about the correct number of training steps that the training script will run and messup the epoch calculation since we count on dataset to report correct `__len__`. It is okay for now, but it is more on the user to make sure to specify the correct bucketing frames. If they have videos that do not match the criteria and have enough frames, it should be discarded in dataset preparation phase, so we never reach this case.

In #34, I faced the same issue so what I did was to just make videos that don't match any bucket default to 1 frame (effectively single image). Not sure what the best thing to do is, but I would rather handle these cases during dataset preparation because that makes most sense to me. LMK your thoughts

The change I made in #34 is as follows:

image

training/mochi-1/dataset.py
```diff
@@ -0,0 +1,444 @@
...
+    def _preprocess_video(self, path: Path) -> torch.Tensor:
+        if self.load_tensors:
+            return self._load_preprocessed_latents_and_embeds(path)
+        else:
+            video_reader = decord.VideoReader(uri=path.as_posix())
+            video_num_frames = len(video_reader)
+            nearest_frame_bucket = min(
+                self.frame_buckets, key=lambda x: abs(x - min(video_num_frames, self.max_num_frames))
+            )
+            if video_num_frames < nearest_frame_bucket:
```
a-r-r-o-w, line 360: Same comment goes here

training/mochi-1/prepare_dataset.py
```diff
@@ -0,0 +1,680 @@
+#!/usr/bin/env python3
```
a-r-r-o-w, line 1: It looks mostly similar to our existing script, so I think the changes are all correct due to the minimal diff - very cool!

training/mochi-1/text_to_video_lora.py
```diff
@@ -0,0 +1,948 @@
...
+    transformer.to(accelerator.device, dtype=weight_dtype)
+    if args.gradient_checkpointing:
+        transformer.enable_gradient_checkpointing()
+
+    # now we will add new LoRA weights to the attention layers
+    transformer_lora_config = LoraConfig(
+        r=args.rank,
+        lora_alpha=args.lora_alpha,
+        init_lora_weights=True,
+        target_modules=["to_k", "to_q", "to_v", "to_out.0"],
```
a-r-r-o-w, line 357: Not required at the moment, and this is more from the diffusion training community that regularly finetunes image models, but finetuning certain layers can make a model worse. Typically, you would want to understand which layers do what to the video by removing that layer and trying to run inference. We should try to throughly understand this for CogVideoX and Mochi, and try to find which layers does it make sense to finetune (and provide users this configurability instead of finetuning all layers by default) for say aesthetics, new concept, temporality improvements, stylized effects, etc. Would you be interested in doing this analysis for Mochi, and I can take it up for Cog?

training/mochi-1/text_to_video_lora.py
```diff
@@ -0,0 +1,948 @@
...
+
+    # For DeepSpeed training
+    model_config = transformer.module.config if hasattr(transformer, "module") else transformer.config
+
+    if args.load_tensors:
+        gc.collect()
+        torch.cuda.empty_cache()
+        torch.cuda.synchronize(accelerator.device)
+
+    def get_sigmas(timesteps, n_dim=4, dtype=torch.float32):
```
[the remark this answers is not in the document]
a-r-r-o-w: I think we'll have to take a deeper look into this soon, because the current training runs seem to not have worked as output videos are random noise.

I'll try and take a better look soon as well!

training/mochi-1/text_to_video_lora.py
```diff
@@ -0,0 +1,948 @@
...
+                    pipe.enable_model_cpu_offload()
+
+                validation_prompts = args.validation_prompt.split(args.validation_prompt_separator)
+                for validation_prompt in validation_prompts:
+                    pipeline_args = {
+                        "prompt": validation_prompt,
+                        "guidance_scale": 4.5,
+                        "height": args.height,
+                        "width": args.width,
+                        "max_sequence_length": 256,
```
a-r-r-o-w, line 825: Just wanted to understand why this change is required, or if we could make it configurable to limit amount of padding tokens. IMO, with the mochi attention mask fix by @DN6, the default 512 should work okay as well

a-r-r-o-w (commented):
Let's jam on this and make the script work asap! Tysm for this

training/mochi-1/dataset.py
sayakpaul: Ah that makes sense. I understand this concern because it resulted in an unexpected error during my trial runs too. I will use your changes in this branch. Thanks!

training/mochi-1/text_to_video_lora.py
sayakpaul: Later yes. But we should try to first find something that is the simplest yet reasonable. But I will make it configurable right now.

training/mochi-1/text_to_video_lora.py
sayakpaul: 512 is not the default for Mochi-1 anywhere.
https://github.com/huggingface/diffusers/blob/acf479bded1ae9fc77b2dc7f316be099abf379ce/src/diffusers/pipelines/mochi/pipeline_mochi.py#L495

https://github.com/genmoai/models/blob/1a96b5df0c83018fcbe7aebbbb1b34781e179921/src/genmo/mochi_preview/pipelines.py#L64

What am I missing?

training/mochi-1/text_to_video_lora.py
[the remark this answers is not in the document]
sayakpaul: Yes, exactly. I am looking into it too.

training/mochi-1/dataset.py
sayakpaul: Hmm, I think with this change, additionally, one needs to also specify the frame buckets to be `[1, 84]`? Because, otherwise, we will have `self.buckets={(84, 480, 848): []}` which will ultimately lead to a `KeyNotFoundError` later during bucket sampling?

training/mochi-1/prepare_dataset.py
```diff
@@ -0,0 +1,681 @@
...
+    video_latents: Optional[torch.Tensor] = None,
+    prompts: Optional[List[str]] = None,
+    prompt_embeds: Optional[torch.Tensor] = None,
+    prompt_attention_mask: Optional[torch.Tensor] = None
+) -> None:
+    metadata = []
+    for i in range(videos.size(0)):
+        video = videos[i:i+1]
+        metadata_dict = {"num_frames": video.size(1), "height": video.size(3), "width": video.size(4)}
+        metadata.append(metadata_dict)
```
sayakpaul, line 328: @a-r-r-o-w take note of this. Otherwise, I don't think the unpacking is proper especially when we're considering a batch size of > 1. LMK if I am missing something.

training/mochi-1/args.py
```diff
@@ -0,0 +1,481 @@
...
+        default=64,
+        help="The lora_alpha to compute scaling factor (lora_alpha / rank) for LoRA matrices.",
+    )
+    parser.add_argument(
+        "--target_modules",
+        nargs="+",
+        type=str,
+        default=["to_k", "to_q", "to_v", "to_out.0"],
+        help="Target modules to train LoRA for."
+    )
```
sayakpaul, line 160: @a-r-r-o-w made the modules configurable, should be okay now?

training/mochi-1/prepare_dataset.sh
```diff
@@ -0,0 +1,49 @@
...
+NUM_GPUS=1
+
+# For more details on the expected data format, please refer to the README.
+DATA_ROOT="/home/sayak/cogvideox-factory/video-dataset-disney"  # This needs to be the path to the base directory where your videos are located.
+CAPTION_COLUMN="prompt.txt"
+VIDEO_COLUMN="videos.txt"
+OUTPUT_DIR="/home/sayak/cogvideox-factory/video-dataset-disney/mochi-1/preprocessed-dataset"
+HEIGHT_BUCKETS="480"
+WIDTH_BUCKETS="848"
+FRAME_BUCKETS="1 84"
```
sayakpaul, line 14: @a-r-r-o-w to account for what I mentioned here:
https://github.com/a-r-r-o-w/cogvideox-factory/pull/90#discussion_r1847602195

training/mochi-1/dataset.py
```diff
@@ -0,0 +1,441 @@
...
+            return index
+
+        if self.load_tensors:
+            image_latents, video_latents, prompt_embeds, prompt_attention_mask = self._preprocess_video(self.video_paths[index])
+
+            # This is hardcoded for now.
+            # Output of the VAE encoding is 2 * output_channels and then it's
+            # temporal compression factor is 6. Initially, the VAE encodings will have
+            # 24 latent number of frames. So, if we were to train with a 
+            # max frame size of 84 and frame bucket of [84], we need to have the following logic.
```
sayakpaul, line 130: @a-r-r-o-w

sayakpaul:
@a-r-r-o-w I have pushed a couple of updates that better reuses the existing `dataset.py` module and reduces a significant LoC. I had to do a fresh `prepare_dataset.py`, though, because we have some major differences, mainly regarding to `prompt_attention_mask`. 

I have also addressed some of your other comments.

Will now look into `sigmas`.

training/mochi-1/text_to_video_lora.py
a-r-r-o-w: Yes, my bad please ignore. I had it stuck in mind from somewhere that the correct sequence length is 512, but seems like I misunderstood something

training/mochi-1/prepare_dataset.py
a-r-r-o-w: Not sure I understand completely. Do you mean to say that metadata_dict is incorrect when batch_size > 1?

training/mochi-1/args.py
a-r-r-o-w: Perfect! I will propagate to Cog scripts as well

training/mochi-1/text_to_video_lora.py
```diff
@@ -0,0 +1,949 @@
...
+
+                # Encode videos
+                if not args.load_tensors:
+                    videos = videos.permute(0, 2, 1, 3, 4)  # [B, C, F, H, W]
+                    latent_dist = vae.encode(videos).latent_dist
+                else:
+                    latent_dist = DiagonalGaussianDistribution(videos)
+
+                videos = latent_dist.sample() 
+                videos = videos[:, :vae_in_channels, ...] # to respect `in_channels` for the vae
```
sayakpaul, line 663: Otherwise, the latent channel dim is 24. 

Cc: @a-r-r-o-w

training/mochi-1/prepare_dataset.sh
a-r-r-o-w: Would making 1 the default be more sensible, or do you think better for user to configure this? The change I applied in our other PR took care of this in a similar fashion by always defaulting to 1 if a suitable frame bucket could not be found:

image

training/mochi-1/text_to_video_lora.py
a-r-r-o-w: Huh wow, interesting. Shouldn't latent_dist already segregate the mean and variance into 12 channels each and already give us hte correct dimensions 🤔 Maybe I'm missing something here...

training/mochi-1/prepare_dataset.py
sayakpaul: Yeah. We would expect the length of `metadata` list to be the same as len(videos)`, i.e., `len(metadata) == len(videos)`. Currently, it won't be the case without the changes.

training/mochi-1/prepare_dataset.sh
sayakpaul: The `1, 84` because of what I mentioned here:
https://github.com/a-r-r-o-w/cogvideox-factory/pull/90#discussion_r1847602195

training/mochi-1/text_to_video_lora.py
sayakpaul: You're right! Not sure what I was thinking. I have rectified that in my local branch and running a trial. Let's see. But shouldn't be a difference.

sayakpaul: [9eea656](https://github.com/a-r-r-o-w/cogvideox-factory/pull/90/commits/9eea656fa07c2924b76cead59916806f4f2ebfb4) should have fixed this.

sayakpaul:
@a-r-r-o-w I am getting a 

```bash
Original Traceback (most recent call last):
  File "/home/sayak/.pyenv/versions/3.10.12/envs/diffusers/lib/python3.10/site-packages/torch/utils/data/_utils/worker.py", line 351, in _worker_loop
    data = fetcher.fetch(index)  # type: ignore[possibly-undefined]
  File "/home/sayak/.pyenv/versions/3.10.12/envs/diffusers/lib/python3.10/site-packages/torch/utils/data/_utils/fetch.py", line 55, in fetch
    return self.collate_fn(data)
  File "/home/sayak/cogvideox-factory/training/mochi-1/text_to_video_lora.py", line 204, in __call__
    videos = torch.stack(videos).to(dtype=self.weight_dtype, non_blocking=True)
```

Is this expected? Do you remember doing something extra to tackle the cases where the number frames was reduced to 1 because frame buckets weren't found?

sayakpaul:
I decided to pad videos having lower number of frames than the nearest frame bucket.

TrickyBarrel:
@sayakpaul Did you manage to find a fix for the random noise in output videos ?

sayakpaul:
Yeah. Will push my updates soon

training/mochi-1/args.py
```diff
@@ -0,0 +1,263 @@
...
+    parser.add_argument(
+        "--cast_dit",
+        action="store_true",
+        help="If we should cast DiT params to a lower precision.",
+    )
+    parser.add_argument(
+        "--compile_dit",
+        action="store_true",
+        help="If we should cast DiT params to a lower precision.",
```
a-r-r-o-w, line 45: The help message here seems out of sync. What does this flag do btw? Because if it's `torch.compile`, we can leverage accelerate for that, no?

training/mochi-1/README.md
```diff
@@ -0,0 +1,97 @@
...
+You can disable intermediate validation by:
+
+```diff
+- --validation_prompt "..." \
+- --validation_prompt_separator ::: \
+- --num_validation_videos 1 \
+- --validation_epochs 1 \
+```
+
+We haven't rigorously tested but without validation enabled, this script should run under 40GBs of GPU VRAM.
```
a-r-r-o-w, line 55: This is for how many frames btw?

training/mochi-1/train.sh
```diff
@@ -0,0 +1,37 @@
...
+export NCCL_P2P_DISABLE=1
+export TORCH_NCCL_ENABLE_MONITORING=0
+
+GPU_IDS="0"
+
+DATA_ROOT="videos_prepared"
+MODEL="genmo/mochi-1-preview"
+OUTPUT_PATH="mochi-lora"
+
+cmd="CUDA_VISIBLE_DEVICES=$GPU_IDS python text_to_video_lora_simple.py \
```
a-r-r-o-w, line 11:
```suggestion
cmd="CUDA_VISIBLE_DEVICES=$GPU_IDS python text_to_video_lora.py \
```

training/mochi-1/text_to_video_lora.py
```diff
@@ -0,0 +1,538 @@
...
+    model_description = f"""
+# Mochi-1 Preview LoRA Finetune
+
+<Gallery />
+
+## Model description
+
+This is a lora finetune of the Moch-1 preview model `{base_model}`.
+
+The model was trained using [CogVideoX Factory](https://github.com/a-r-r-o-w/cogvideox-factory) - a repository containing memory-optimized training scripts for the CogVideoX, Mochi family of models using [TorchAO](https://github.com/pytorch/ao) and [DeepSpeed](https://github.com/microsoft/DeepSpeed). The scripts were adopted from [CogVideoX Diffusers trainer](https://github.com/huggingface/diffusers/blob/main/examples/cogvideo/train_cogvideox_lora.py).
```
a-r-r-o-w, line 94:
```suggestion
The model was trained using [CogVideoX Factory](https://github.com/a-r-r-o-w/cogvideox-factory) - a repository containing memory-optimized training scripts for the CogVideoX and Mochi family of models using [TorchAO](https://github.com/pytorch/ao) and [DeepSpeed](https://github.com/microsoft/DeepSpeed). The scripts were adopted from [CogVideoX Diffusers trainer](https://github.com/huggingface/diffusers/blob/main/examples/cogvideo/train_cogvideox_lora.py).
```

training/mochi-1/text_to_video_lora.py
```diff
@@ -0,0 +1,538 @@
...
+            )
+
+    model_description = f"""
+# Mochi-1 Preview LoRA Finetune
+
+<Gallery />
+
+## Model description
+
+This is a lora finetune of the Moch-1 preview model `{base_model}`.
```
a-r-r-o-w, line 92:
```suggestion
This is a lora finetune of the Mochi-1 preview model `{base_model}`.
```

training/mochi-1/text_to_video_lora.py
```diff
@@ -0,0 +1,538 @@
...
+## Download model
+
+[Download LoRA]({repo_id}/tree/main) in the Files & Versions tab.
+
+## Usage
+
+Requires the [🧨 Diffusers library](https://github.com/huggingface/diffusers) installed.
+
+```py
+TODO
```
a-r-r-o-w, line 105: I think the same code from README could be put here

training/mochi-1/text_to_video_lora.py
```diff
@@ -0,0 +1,538 @@
...
+    )
+
+    transformer.requires_grad_(False)
+    transformer.to("cuda")
+    if args.gradient_checkpointing:
+        transformer.enable_gradient_checkpointing()
+    if args.cast_dit:
+        transformer = cast_dit(transformer, torch.bfloat16)
+    if args.compile_dit:
+        transformer.compile()
```
a-r-r-o-w, line 264: Interesting... I did not know we had a method that allows compilation this way. This is torch.compile-ation right?

a-r-r-o-w (approved):
Thanks Sayak! Just some minor comments and we should be good to merge

We can work on the refactors and generic-fication when we find the time later on

a-r-r-o-w:
Also could you point me to the logs from your most promising run so far?

sayakpaul:
> Also could you point me to the logs from your most promising run so far?

https://wandb.ai/sayakpaul/mochi-1-lora/runs/hu344t6o

Here's one run with the original trainer but on the same dataset:
https://wandb.ai/sayakpaul/mochi-1-lora/runs/de4iwnvf

We can see that the loss dynamics are the same. Additionally, here's a sample derived intermediate with the original trainer:

https://github.com/user-attachments/assets/1163b82c-10bb-4399-9b1d-1bf0e3badcda

The one we see here is not too bad: https://wandb.ai/sayakpaul/mochi-1-lora/runs/hu344t6o. Of course, I suspect the other quality issues will go away once https://github.com/huggingface/diffusers/pull/10033 is merged. But there is a blocker with that: https://huggingface.slack.com/archives/C065E480NN9/p1732808679447279?thread_ts=1732688413.727359&cid=C065E480NN9. 

LMK if you have any other questions.

training/mochi-1/text_to_video_lora.py
sayakpaul: Yes sir.

training/mochi-1/text_to_video_lora.py
sayakpaul: Done in e1866d8.

training/mochi-1/README.md
sayakpaul: 37 frames (similar to the original example).

training/mochi-1/args.py
sayakpaul: Good catch. Fixed in 9c86706. 

Since we're not leveraging `accelerate` at the moment, I decided to expose it separately.

a-r-r-o-w: Oh I missed that we're not using accelerate here. Cool, we'll support it later

a-r-r-o-w:
Wow, these look very promising! The quality will definitely improve once Dhruv's PR is in, both for training and inference. I think we're good to merge then

sayakpaul:
@a-r-r-o-w thanks! Do you think it'd be prudent to get https://github.com/huggingface/diffusers/pull/10031/ in, as it was also critical?

PR #98, opened 2024-11-27
docs: fix help message in args.py

Leojc:


training/args.py
```diff
@@ -78,7 +78,7 @@ def _get_dataset_args(parser: argparse.ArgumentParser) -> None:
         nargs="+",
         type=int,
         default=[49],
-        help="CogVideoX1.5 need to guarantee that ((num_frames - 1) // self.vae_scale_factor_temporal + 1) % patch_size_t != 0, such as 53"
+        help="CogVideoX1.5 need to guarantee that ((num_frames - 1) // self.vae_scale_factor_temporal + 1) % patch_size_t == 0, such as 53"
```
a-r-r-o-w, line 81:
```suggestion
        help="For CogVideoX 1.0, this can be any value of the form 4 * k + 1 (such as 49) for training/inference. For CogVideoX 1.5, this can any value of the form 8 * k - 3 (such as 85) for training, but in inference, we can use a trick which allows 4 * k + 1 frames to work as well"
```

a-r-r-o-w (approved):
Just a small change to make it clear

PR #103, opened 2024-11-29
Multi-resolution training

BuffMcBigHuge:
This allows for training at varied resolutions, such as portrait, with 1.5 I2V.

### DIFF
--- a/training/args.py
+++ b/training/args.py
@@ -81,6 +81,18 @@ def _get_dataset_args(parser: argparse.ArgumentParser) -> None:
         help="CogVideoX1.5 need to guarantee that ((num_frames - 1) // self.vae_scale_factor_temporal + 1) % patch_size_t == 0, such as 53"
     )
     parser.add_argument(
+        "--base_height",
+        type=int,
+        default=480,
+        help="Base height for rotary positional embeddings",
+    )
+    parser.add_argument(
+        "--base_width",
+        type=int,
+        default=720,
+        help="Base width for rotary positional embeddings",
+    )
+    parser.add_argument(
         "--load_tensors",
         action="store_true",
         help="Whether to use a pre-encoded tensor dataset of latents and prompt embeddings instead of videos and text prompts. The expected format is that saved by running the `prepare_dataset.py` script.",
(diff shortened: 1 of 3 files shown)


training/args.py
```diff
@@ -80,6 +80,18 @@ def _get_dataset_args(parser: argparse.ArgumentParser) -> None:
         default=[49],
         help="CogVideoX1.5 need to guarantee that ((num_frames - 1) // self.vae_scale_factor_temporal + 1) % patch_size_t == 0, such as 53"
     )
+    parser.add_argument(
+        "--base_height",
+        type=int,
+        default=480,
+        help="Base height for rotary positional embeddings",
+    )
```
a-r-r-o-w, line 88: Could we use the value from transformer config instead of making this a parameter? base_height could be something like `model_config.sample_height * 8`, and likewise for width. This is because it's not easy for users to know beforehand what to specify here. The 8 comes from vae spatial compression ratio so that could be made a separate variable for clarity if you'd like

Leojc:
Maybe add these args to the prepare_rotary_positional_embeddings [link](https://github.com/a-r-r-o-w/cogvideox-factory/blob/76a9f2b9111f5a9d9feb5fabefda3fb7fe68bf7b/training/cogvideox_image_to_video_lora.py#L784) according to the Pipeline:
```python
base_height=model_config.sample_height * VAE_SCALE_FACTOR_SPATIAL,
base_width=model_config.sample_width * VAE_SCALE_FACTOR_SPATIAL,
```

Leojc:
> Maybe add these args to the prepare_rotary_positional_embeddings [link](https://github.com/a-r-r-o-w/cogvideox-factory/blob/76a9f2b9111f5a9d9feb5fabefda3fb7fe68bf7b/training/cogvideox_image_to_video_lora.py#L784) according to the Pipeline:
> 
> ```python
> base_height=model_config.sample_height * VAE_SCALE_FACTOR_SPATIAL,
> base_width=model_config.sample_width * VAE_SCALE_FACTOR_SPATIAL,
> ```

Oh... It has been done in this pr #107

BuffMcBigHuge:
Thanks @Leojc

PR #107, opened 2024-12-01
RoPE fixes for 1.5, bfloat16 support in prepare_dataset, gradient_accumulation grad norm undefined fix

a-r-r-o-w:


training/cogvideox_image_to_video_lora.py
```diff
@@ -398,6 +398,8 @@ def main(args):
 
     VAE_SCALING_FACTOR = vae.config.scaling_factor
     VAE_SCALE_FACTOR_SPATIAL = 2 ** (len(vae.config.block_out_channels) - 1)
+    RoPE_BASE_HEIGHT = transformer.config.sample_height * VAE_SCALE_FACTOR_SPATIAL
+    RoPE_BASE_WIDTH = transformer.config.sample_width * VAE_SCALE_FACTOR_SPATIAL
```
sayakpaul, line 402: In case of DeepSpeed we won't be able to access `transformer.config`. Let's use `model_config`?

training/cogvideox_image_to_video_lora.py
```diff
@@ -828,6 +833,10 @@ def load_model_hook(models, input_dir):
                     gradient_norm_before_clip = get_gradient_norm(transformer.parameters())
                     accelerator.clip_grad_norm_(transformer.parameters(), args.max_grad_norm)
                     gradient_norm_after_clip = get_gradient_norm(transformer.parameters())
+                    logs.update({
+                        "gradient_norm_before_clip": gradient_norm_before_clip,
+                        "gradient_norm_after_clip": gradient_norm_after_clip,
+                    })
```
sayakpaul, line 839: `ruff` should format it like so:

```py
logs.update(
    {
        "gradient_norm_before_clip": gradient_norm_before_clip,
        "gradient_norm_after_clip": gradient_norm_after_clip,
    }
)
```

training/cogvideox_image_to_video_lora.py
```diff
@@ -876,15 +885,10 @@ def load_model_hook(models, input_dir):
...
-                logs.update(
-                    {
-                        "gradient_norm_before_clip": gradient_norm_before_clip,
-                        "gradient_norm_after_clip": gradient_norm_after_clip,
-                    }
-                )
+            logs.update({
+                "loss": loss.detach().item(),
+                "lr": last_lr,
+            })
```
sayakpaul, line 891:
```suggestion
            logs.update({"loss": loss.detach().item(), "lr": last_lr})
```

LoC 119 limit should be fine with it.

training/cogvideox_text_to_video_lora.py
```diff
@@ -328,6 +328,8 @@ def main(args):
 
     VAE_SCALING_FACTOR = vae.config.scaling_factor
     VAE_SCALE_FACTOR_SPATIAL = 2 ** (len(vae.config.block_out_channels) - 1)
+    RoPE_BASE_HEIGHT = transformer.config.sample_height * VAE_SCALE_FACTOR_SPATIAL
+    RoPE_BASE_WIDTH = transformer.config.sample_width * VAE_SCALE_FACTOR_SPATIAL
```
sayakpaul, line 332: Same as above.

training/cogvideox_text_to_video_lora.py
```diff
@@ -736,6 +741,10 @@ def load_model_hook(models, input_dir):
                     gradient_norm_before_clip = get_gradient_norm(transformer.parameters())
                     accelerator.clip_grad_norm_(transformer.parameters(), args.max_grad_norm)
                     gradient_norm_after_clip = get_gradient_norm(transformer.parameters())
+                    logs.update({
+                        "gradient_norm_before_clip": gradient_norm_before_clip,
+                        "gradient_norm_after_clip": gradient_norm_after_clip,
+                    })
```
sayakpaul, line 747: Same as above.

training/cogvideox_image_to_video_lora.py
```diff
@@ -828,6 +833,10 @@ def load_model_hook(models, input_dir):
                     gradient_norm_before_clip = get_gradient_norm(transformer.parameters())
                     accelerator.clip_grad_norm_(transformer.parameters(), args.max_grad_norm)
                     gradient_norm_after_clip = get_gradient_norm(transformer.parameters())
+                    logs.update({
```
sayakpaul, line 836: Should we not consider if we're using DeepSpeed and log it accordingly?

I referring to the following comment:

```
# gradnorm + deepspeed: https://github.com/microsoft/DeepSpeed/issues/4555
```

training/cogvideox_text_to_video_lora.py
```diff
@@ -776,15 +785,10 @@ def load_model_hook(models, input_dir):
...
-                logs.update(
-                    {
-                        "gradient_norm_before_clip": gradient_norm_before_clip,
-                        "gradient_norm_after_clip": gradient_norm_after_clip,
-                    }
-                )
+            logs.update({
+                "loss": loss.detach().item(),
+                "lr": last_lr,
+            })
```
sayakpaul, line 791: Same as above.

training/prepare_dataset.py
```diff
@@ -287,11 +287,13 @@ def compute_prompt_embeddings(
 
 
 def save_image(image: torch.Tensor, path: pathlib.Path) -> None:
-    image = to_pil_image(image)
+    image = image.to(dtype=torch.float32).clamp(-1, 1)
+    image = to_pil_image(image.float())
     image.save(path)
```
sayakpaul, line 292: Could you explain this change?

sayakpaul (commented):
Thank you!

training/cogvideox_image_to_video_lora.py
a-r-r-o-w: We can access it as this point because it is still not wrapped by deepspeed

training/cogvideox_image_to_video_lora.py
a-r-r-o-w: we cannot calculate grad_norm for deepspeed, so this will be None

training/prepare_dataset.py
a-r-r-o-w: F.ToPILImage does not support bfloat16. So if someone was to do precomputation in bfloat16 and try to save the images/videos in viewable format, it would error out. casting to fp32 fixes this

sayakpaul (approved):

training/cogvideox_image_to_video_lora.py
a-r-r-o-w: formatted

PR #108, opened 2024-12-02
fix: i2v launch script default prompts.txt path

bruce2233:
matching preprocess output format


a-r-r-o-w:
Sorry for not being able to review in the past. We support automatic detection of multiple dataset formats and this should be handled now. Thanks for your interest!

PR #129, opened 2024-12-20
Precomputation of conditions and latents

a-r-r-o-w:


sayakpaul:
Actually I think I have an idea which I would like to consider before this is merged. I am on the move currently but get the comments in by today before evening.

a-r-r-o-w:
Sounds good, LMK when you're free!

finetrainers/trainer.py
```diff
@@ -364,34 +578,46 @@ def train(self) -> None:
...
+                            text_encoder=self.text_encoder,
+                            tokenizer_2=self.tokenizer_2,
+                            text_encoder_2=self.text_encoder_2,
+                            prompt=prompts,
+                            device=accelerator.device,
+                            dtype=weight_dtype,
+                        )
+                    else:
+                        latent_conditions = batch["latent_conditions"]
+                        other_conditions = batch["other_conditions"]
```
sayakpaul, line 610: `other_conditions` seem a bit unintuitive in this context. Prefer explicit naming.

finetrainers/utils/torch_utils.py
```diff
@@ -6,3 +9,21 @@ def unwrap_model(accelerator: Accelerator, model):
...
+        if device is not None:
+            x = x.to(device)
+        if dtype is not None:
+            x = x.to(dtype)
+    elif isinstance(x, dict):
+        if device is not None:
+            x = {k: align_device_and_dtype(v, device, dtype) for k, v in x.items()}
+        if dtype is not None:
+            x = {k: align_device_and_dtype(v, device, dtype) for k, v in x.items()}
+    return x
```
sayakpaul, line 29: I think this is an overkill given it's just iterating over a dict and performing `to()`. IMO, an in-line `to()` doesn't necessarily hurt even if it contributes to increased LoC.

sayakpaul (changes requested):

README.md
```diff
@@ -143,6 +143,110 @@ video = pipe("<my-awesome-prompt>").frames[0]
 export_to_video(video, "output.mp4", fps=8)
 ```
 
+### Memory Usage
+
+LoRA with rank 128, batch size 1, gradient checkpointing, optimizer adamw, `49x512x768` resolution, **without precomputation**:
```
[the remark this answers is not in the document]
a-r-r-o-w: Okay

finetrainers/hunyuan_video/hunyuan_video_lora.py
```diff
@@ -35,24 +34,41 @@ def load_components(
...
+    **kwargs,
+) -> Dict[str, nn.Module]:
+    transformer = HunyuanVideoTransformer3DModel.from_pretrained(
+        model_id, subfolder="transformer", torch_dtype=transformer_dtype, revision=revision, cache_dir=cache_dir
+    )
+    scheduler = FlowMatchEulerDiscreteScheduler()
+    return {"transformer": transformer, "scheduler": scheduler}
+
+
 def initialize_pipeline(
```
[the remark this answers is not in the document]
a-r-r-o-w: Does not really require any changes. All the currently loaded models will be passed in (if no precomputation, then all text encoders, vae and transformer. if precomputation, then only transformer). The remaining models that are not provided will be loaded with defaults because they are all needed for inference.

One thing I still have in mind is implementing precomputation for validation prompts, but that's something I will revisit in a future PR because I'm busy with a few other things on diffusers-side that I would like to work on first

finetrainers/ltx_video/ltx_video_lora.py
```diff
@@ -12,32 +12,45 @@
...
+    return {"vae": vae}
+
+
+def load_diffusion_models(
+    model_id: str = "Lightricks/LTX-Video",
+    transformer_dtype: torch.dtype = torch.bfloat16,
+    revision: Optional[str] = None,
+    cache_dir: Optional[str] = None,
+    **kwargs,
+) -> Dict[str, nn.Module]:
```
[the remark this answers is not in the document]
a-r-r-o-w: Happy to correct this but as a general rule of thumb, let's not worry too much about things that don't matter at this stage of the trainer. We should focus on getting the features out more quickly based on feedback from the community and people that are actually using the project (The Banadoco discord community is one good example where I've made some friends to gather feedback from). Faster iteration cycles and quick feature PRs and support for new things should be our goal, WDYT?

finetrainers/ltx_video/ltx_video_lora.py
```diff
@@ -114,19 +127,52 @@ def prepare_latents(
...
+        latents = _pack_latents(latents, patch_size, patch_size_t)
+        return {"latents": latents, "num_frames": num_frames, "height": height, "width": width}
+    else:
+        if vae.use_slicing and image_or_video.shape[0] > 1:
+            encoded_slices = [vae._encode(x_slice) for x_slice in image_or_video.split(1)]
+            h = torch.cat(encoded_slices)
+        else:
+            h = vae._encode(image_or_video)
+        _, _, num_frames, height, width = h.shape
+        # TODO(aryan): this is very very very stupid, but anything to make it work for now. refactor and design better later
```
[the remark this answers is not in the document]
a-r-r-o-w: Yes sorry I should have been more clear.

So, we are returning vae latents_mean and latents_std here because we need it for latent normalization in `post_latent_preparation`. But that method is not called until training starts and a DiagonalGaussianDistribution is created. In our precomputation, we also end up saving the latents_mean and latents_std per latent file, which can quickly add up to use a lot of disk storage for bigger precomputation runs (for lora, it doesn't matter too much because it's only about < 100 videos most of the time). Ideally, we should only save this once but since I didn't find time to look into implementing that yet, I just let it be the way it is

finetrainers/trainer.py
```diff
@@ -111,36 +176,189 @@ def prepare_models(self) -> None:
...
+                pin_memory=self.args.pin_memory,
+            )
+            return
+
+        logger.info("Precomputed conditions and latents not found. Running precomputation.")
+
+        # At this point, no models are loaded, so we need to load and precompute conditions and latents
+        condition_components = self.model_config["load_condition_models"](**self._get_load_components_kwargs())
+        self._set_components(condition_components)
+        self._move_components_to_device()
```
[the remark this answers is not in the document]
a-r-r-o-w: Because we have to move the modeling components from CPU to GPU, no? It's a handy helper method that moves all currently loaded modules (whether in precomputation stage or training stage) to the GPU

finetrainers/trainer.py
```diff
@@ -111,36 +176,189 @@ def prepare_models(self) -> None:
...
+
+        if self.vae is not None:
+            if self.args.enable_slicing:
+                self.vae.enable_slicing()
+            if self.args.enable_tiling:
+                self.vae.enable_tiling()
+
+        progress_bar = tqdm(
+            range(0, len(self.dataset)),
+            desc="Precomputing latents",
```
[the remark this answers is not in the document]
a-r-r-o-w: We can look into condition precomputation part and design for it later. What we should aim for is:
- Text encoding related parts don't load VAE
- VAE related parts don't load text encoder

So what we now call "precomputing latents" is basically "prepare all things related to VAE", and "precomputing conditions" is basically "prepare all things related to text encoders". Additionally, we might have other things like control condition preparations, which would go other VAE-related stuff. Any other kind of embeddings, like facial image embeds, can go into a separate dedicated method that does so in a way that doesn't simultaneously load text encoder or vae at the same time. We should also ensure that the condition preparation methods should be independent of accelerator states, so that in the future we can implement FSDP and tensor parallelism without having to refactor.

>  Later, we could also make it so that each of the above is executed on different GPUs should a multi-GPU configuration is discovered (and based on user preferences).

Currently, we first precompute all embeddings distributed across GPUs first, and then precompute all latents across all GPUs. Is this not how we want it?

finetrainers/trainer.py
```diff
@@ -158,12 +376,8 @@ def prepare_trainable_parameters(self) -> None:
 
         # TODO(aryan): handle torch dtype from accelerator vs model dtype; refactor
         self.state.weight_dtype = weight_dtype
-        self.text_encoder.to(self.state.accelerator.device, dtype=weight_dtype)
-        self.transformer.to(self.state.accelerator.device, dtype=weight_dtype)
-        self.vae.to(self.state.accelerator.device, dtype=weight_dtype)
-
-        if self.text_encoder_2 is not None:
-            self.text_encoder_2.to(self.state.accelerator.device, dtype=weight_dtype)
```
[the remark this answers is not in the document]
a-r-r-o-w: I've added a helper method `_move_components_to_device` that moves all currently loaded models to accelerator's device

finetrainers/trainer.py
a-r-r-o-w: What would you prefer? `text_conditions` works?

finetrainers/utils/torch_utils.py
a-r-r-o-w: Not really, this will be hard to handle and create a mess of if-statements. We don't beforehand know all the kinds of prepared conditions by name. I mean, we do because we will be implementing it but there can be a lot of different names in the dictionaries + this would allow any custom implementations from users to directly benefit from just adding any more amount of tensors in the conditions they want and it would "just work". We can have dictionary keys as prompt_embeds, pooled_prompt_embeds, latents, control_latents, etc. to name a few, as well as single tensors and list of tensors nested in dictionaries, and many more as we move forward. Having an if-statement to check if a key exists and then move it to correct device would not be very nice

finetrainers/ltx_video/ltx_video_lora.py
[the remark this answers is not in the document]
sayakpaul: We can always obtain `latents_mean` and `latents_std` from the VAE config file, right? And loading a config file is very easy with `load_config`. Should be a relatively easy change.

finetrainers/ltx_video/ltx_video_lora.py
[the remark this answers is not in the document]
sayakpaul: Why bother about type-annotations at all then :D An incorrect type-annotation is more confusing than not having one. Remember that code is also ready by folks who tend to become the power-users of a project. Since the project codebase is relatively manageable at this stage, I think this is okay.

finetrainers/ltx_video/ltx_video_lora.py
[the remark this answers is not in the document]
a-r-r-o-w: It's not easy in this case, which is why I implemented it the way it is 😅 The mean and std are per channel tensors of shape (128,) part of the state dict. Maybe I can just copy the values and paste here as a list. Would that work?

finetrainers/trainer.py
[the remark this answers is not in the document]
sayakpaul: Okay. Computing captions are independent from computing latents. So, this can happen parallelly. But no biggie.

finetrainers/ltx_video/ltx_video_lora.py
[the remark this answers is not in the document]
sayakpaul: I can look into it then. Not all the VAEs would have std and mean so, that will also need to be conditioned accordingly. Leave [this comment](https://github.com/a-r-r-o-w/finetrainers/pull/129#discussion_r1895532692) as a note, then.

finetrainers/trainer.py
sayakpaul: `text_conditions` in case when we're dealing them and `conditional_image_latents` when it's image-to-video.

finetrainers/ltx_video/ltx_video_lora.py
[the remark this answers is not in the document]
sayakpaul: I expected [this VAE config](https://huggingface.co/Lightricks/LTX-Video/blob/main/vae/config.json) to have `latents_std` and `latents_mean`, following what we do in others. Why does it not have things? 

> Is it okay if I copy the values from the state dict and directly paste into the file?

Yeah okay with a note.

a-r-r-o-w:
> any kind of precomputation is treated as "data processing" and is separated from trainer classes.

I respectfully disagree. I may not have the experience you have, but as far as our initial goals were aimed towards, we wanted to make video model training more accessible, not target large-scale training. Adding extra steps and more things that users need to follow instructions for it to work a big no-go IMO. Yes, we can eventually try and be more ambitious but for now, we should focus on streamlining things as much as possible and supporting all things lora (normal lora, control lora, turbo distillation loras, etc.)

From some feedback I gathered, people who currently use this project want two things:
- Create a videos.txt and prompts.txt file
- Click some fancy switches and knobs to tweak hyperparameters with, and press a button to start finetuning with.

Ofcourse, this is not a UI, so the button part is not to be considered, and is a battle for someone else ot fight. But opt-in precomputation like we do here from just two text files looks ideal to me (for now). We can reconsider in future as you mention.

Me and you know about things and can work our way through all this, but I want you to think in the way of a general user with no programming knowledge, who is just experimenting from a creator background, has read some tutorials online and just wants to use the sensible defaults, that they will find all of this to be non-sense. Just from the history of diffusion inference in the past two years, I believe there's a lot of lessons to be learnt on what got most popular among users who actually use these kinds of tools daily.

Large scale training is quite a ways away from what we have. Basic data parallelism like we have will never scale IMO, so first, that is what we should tackle before designing anything else if that's what we want to prioritize.

> Functional APIs for each of the pipelines that performs that necessary precomputions.

Awesome, so this is where we are at currently. We just need to "API"-fy it a bit more so that custom methods/callbacks can be passed in

> This will allow us to have a greater degree of flexibility in how we configure and utilize accelerators based on the compute patterns of the precomputation stage, which can be different from training.

Utilizing accelerators based on compute patterns of precomputation stage affect training...? I don't quite follow what you mean. Maybe we can hop in DM and I can try to understand?

> I also don't like train.py as the main entrypoint to train models given each trainer would be separate in classes and will implement their own atomic modules. Refer to https://huggingface.co/docs/trl/example_overview as an example. For now, train.py is fine, though. In short, each trainer implements an example on its own to keep things very clear and friendly to refer to. Example: https://github.com/huggingface/trl/blob/main/trl/scripts/dpo.py.

I would like to distinguish between two things here:
- Algorithm-specific training (LoRA vs Full finetuning vs Control LoRA vs ControlNet vs Distillation vs etc.)
- Model-specific training (LTXV vs Hunyuan vs Cog vs etc.)

TRL implements what I would call "algorithm-specific" training. It works with different models because many things are common and transformers provides some great abstractions, but also because it is mostly just a single model and not many moving components. The same does not apply to diffusion models, as we have discussed time and again how many moving components are involved, and we will most definitely require model-specific implementations because different models process things differently.

But OTOH, most underlying training algorithms are independent of the modeling architecture itself. For example, with a decent abstraction like what we have now, lora training works for both LTX and Hunyuan without any lora stuff interleaving with the single file implementations we have regarding model loading, condition and latent computation, etc. I believe, similar would be the case for full finetuning, distillation, and pretty much everything else (we can break away into better abstraction when we get to it, but let's not prematurely design it with that in mind because it will only slow us down).

I would prefer doing the following:
- Do thing X
- If it works without modifying current implementation too much or not at all, good. Proceed with it
- If it requires substantial modifications, perform the refactor.

These are my thoughts on how we should proceed to maximize our speed of integrating new models and techniques that we want to support. Premature refactoring might bring us abstractions that look good now, but will slow us down and cause further required refactoring in future.

LMK what you think so we can proceed accordingly.

I will address your concerns shortly and push the relevant changes. Thanks for the reviews!

finetrainers/trainer.py
a-r-r-o-w: I will go with `text_conditions` and `latent_conditions` for the moment. We don't support image-to-video at the moment, but I would maybe combine image/control all under a single name because they will effectively just be latents of some kind

sayakpaul:
> Utilizing accelerators based on compute patterns of precomputation stage affect training...? I don't quite follow what you mean. Maybe we can hop in DM and I can try to understand?

I didn't say "affect". I am saying it gives us a chance to better utilize accelerators in the way we want to. Accelerator distribution for precomputation isn't always necessarily the same as training. Below is an example: 

* Use first the GPU for text embeddings.
* The rest of the two for latent computation.

(both happening in parallel)

This (the way accelerators are being utilized) is not what we do during training. 

But okay to not proceed in the direction I mentioned. But I won't absolutely prefer having a single `Trainer` class if that is what you absolutely want. For every model, I do envision a separate `Trainer` class because maintenance-wise, it's better and easier to maintain. And if we cannot agree on this design, then this needs further discussion.

a-r-r-o-w:
> But okay to not proceed in the direction I mentioned. But I won't absolutely prefer having a single Trainer class if that is what you absolutely want

What is okay with me is:
- Separate class per algorithm. So one for lora, one for full finetuning, one for task X, one for task Y, etc.

What is not okay with me:
- One for LTX LoRA, one for LTX Full finetuning, one for Mochi Control LoRA, one for Mochi normal LoRA, etc. This will be a mess

Preparing conditions and latents is a simple enough thing IME that won't require an implementation for each kind of algorithm and can generally be re-used.

I would also not be okay if each trainer class had to implement their own N different "algorithm" implementations. Basically, I don't want LTXTrainer to have its own implementation for all the different mentioned tasks unless absolutely necessary. Doing so will only create bugs and make it harder to maintain. If it were rather called something like `LTXTrainingPipeline` which implemented all the necessary helper methods, it would be better for me - these should have absolutely no training related things intermixed with it and should be completely decoupled. The one thing that would be okay to not decouple from training would be loss computation, because we know there are two broad categories of models that exist today from diffusion perspective.

What would be more likeable to me is `LoRATrainer`, `FullTrainer`, `GuidanceDistillationTrainer`, etc. All the other underlying parts from model-specific perspective can be re-used across this IMO.

sayakpaul:
So, this means you're not okay with the direction taken in https://github.com/a-r-r-o-w/finetrainers/pull/130, right?

a-r-r-o-w:
> Very minimal differences between LoRATrainer and `FullTrainer, though. So, I don't see the point.

Okay, I feel like despite my efforts, it is extremely hard to communicate my point across. Please make an effort to not pick two of the simplest examples and extrapolate it to the entirety of the message. Do we not agree that LoRA is a different training technique than full finetuning? I am trying to discuss about training algorithms, not how similar/dissimilar they are.

My point is, we should have trainers specific to algorithms. Not models. It should be model-independant as long as the model provides all the helpful utility functions required for training. Let's not tie in the two together - this inevitably leads to bugs and I've seen my fair share with the diffusers scripts that do it model-specifically. It is hard to use and unstable, but it serves its purpose in that it allows for hacking and exploration.

Here are the directions I'm not okay with in #130:
- Dataset preparation tied into every trainer. I don't understand why this would be required. Every model takes in the same format of data for a given task. Happy to understand from an example that shows otherwise
- Trainable parameter preparation tied into every trainer. This should be done at one level higher up the abstraction chain depending on the task being trained irrespective of the model used unless absolutely required.
- Loss calculation. There are only two kinds of losses with diffusion models in regards to lora training, which is all we support for now. DDPM and Flow matching. Anything else should be an augmentation loss, which we could be used in any other model. These should be enabled by training flags. However, we are not at the stage where we could consider things like this (for example, motion regularization or latent perceptual loss), so it is okay for now. Unless a paper specifically introduces a loss for a model, and that loss is not compatible with other models, we should use the normal base losses.

To reiterate, the models should not tie in with the actual training algorithm in any way. In all trainers I've looked at, those that do this have very complicated code (but partially also because they build atop the LDM codebase). SimpleTuner is a good example that does not tie in the loss to the model itself, and to me is the most successful training project built on Diffusers because it "just works" OOTB and you don't have to think much to get it to work. I would take inspiration from there and follow some design patterns, but improve wherever possible and customize to our broader set of goals.

sayakpaul:
It's okay to have disagreements and I can see some of your points. So, I will roll with this direction. Making changes later isn't a big deal. 

The reason why I wanted to separate dataset prep was because not all models have the same components we need during precomputation (different text encoders, different masking, etc.). So, unifying that under a single method becomes very messy. 

And then for caption dropout, different models do it differently as mentioned the other day. It's easy to miss this detail but it plays a crucial role, sometimes. 

For loss, Mochi does it differently than the other flow models we have.

a-r-r-o-w:
> The reason why I wanted to separate dataset prep was because not all models have the same components we need during precomputation (different text encoders, different masking, etc.). So, unifying that under a single method becomes very messy.

This, I agree with. What I want is for the utility functions to be part of the model-specific implementations. Not the loading and dataloader creation itself. That should be a separate decoupled component in itself. We leverage the utility functions with conditional in the trainer (just throwing out ideas):

```
text_conditions = self.model_trainer_pipeline.prepare_conditions(...)
# prepare conditions will get a one or more prompts based on some configured batch
# size, and prepare everything that it requires using multiple text encoders if needed.
# For example, I already do this without much complications for Hunyuan which uses
# 2 text encoders
```

> And then for caption dropout, different models do it differently as mentioned the other day. It's easy to miss this detail but it plays a crucial role, sometimes.

Again, this is a trainer-specific detail and should not be tied into the model utilities itself. The trainer will invoke preparation methods. If it's an "empty"-ing dropout, null prompt will be passed to the preparation method. If it's a "tensor zero"-ing dropout, it can be done post-preparation. We already do this. This should be configurable via CLI args to enable exploration of what works, not hardcode it for certain models. We can provide sensible guidelines on what to do per-model.

> For loss, Mochi does it differently than the other flow models we have.

It's one case among the thousands of flow-matching models. It's not a different flow matching algorithm as such. It's just a mistake that occurred in the original codebase, as I like to believe. Instead of the usual (timestep 1000 = noise) and (timestep = clean), they have a model trained with an inverse objective.

> TL;DR -- let's get this PR merged. I will work on Cog next as explained over DM.

Thank you. I realize I might have spoken a bit in spirit and strongly about the way I expect moving forward. I think it is good to have these discussions in early stages before realizing it later. Working my way through the reviews and will push shortly

sayakpaul:
> Again, this is a trainer-specific detail and should not be tied into the model utilities itself. The trainer will invoke preparation methods. If it's an "empty"-ing dropout, null prompt will be passed to the preparation method. If it's a "tensor zero"-ing dropout, it can be done post-preparation. We already do this. This should be configurable via CLI args to enable exploration of what works, not hardcode it for certain models. We can provide sensible guidelines on what to do per-model.

Just so you know it's can be model-specific :D If a model was pre-trained using a particular caption-dropping method it will have to be accounted for. So, we can't eliminate possibility and it's something a user should be mindful about. By default, we shouldn't do any caption dropping out hence. I guess, that is fair?

a-r-r-o-w:
> Just so you know it's can be model-specific :D If a model was pre-trained using a particular caption-dropping method it will have to be accounted for. So, we can't eliminate possibility and it's something a user should be mindful about. By default, we shouldn't do any caption dropping out hence. I guess, that is fair?

Very rarely it's the case that an empty/null prompt is not used. Maybe some of the recent ones, and for these cases, when the trainer object is initialized, I would like a warning be raised than disabling the possibility of doing anything else altogether.

```
# Somewhere in the model utility code
if model_name == "mochi" and args.caption_dropout_technique == "empty":
    logger.warning("You did bad thing. Maybe don't do it.")
```

Just like I said above, let's get to it when we get to it instead of thinking about all this. It only slows us down having pointless debates. None of the models currently require us to diverge from the sensible defaults

sayakpaul:
Okay. I was trying to be thorough about the issues that can silently stem in training but it seems like we don't have to be worried about them that much for now. Let's not call that a pointless debate.

README.md
```diff
@@ -143,6 +143,60 @@ video = pipe("<my-awesome-prompt>").frames[0]
...
+}
+```
+
+| stage                   | memory_allocated | max_memory_reserved |
+|:-----------------------:|:----------------:|:-------------------:|
+| before training start   | 13.486           | 13.879              |
+| before validation start | 14.146           | 17.623              |
+| after validation end    | 14.146           | 17.623              |
+| after epoch 1           | 14.146           | 17.623              |
+| after training end      | 4.461            | 17.623              |
```
sayakpaul, line 169: Very nice!

README.md
```diff
@@ -169,8 +223,7 @@ OUTPUT_DIR="/path/to/models/hunyuan-video/hunyuan-video-loras/hunyuan-video_caki
 
 # Model arguments
 model_cmd="--model_name hunyuan_video \
-  --pretrained_model_name_or_path tencent/HunyuanVideo
-  --revision refs/pr/18"
+  --pretrained_model_name_or_path hunyuanvideo-community/HunyuanVideo"
```
sayakpaul, line 226: Thank god, this is sorted now!

README.md
```diff
@@ -272,10 +325,69 @@ output = pipe(
...
+Note: requires about `47` GB of VRAM with precomputation. If validation is not performed, the memory usage is reduced to about `42` GB.
+
 </details>
 
 If you would like to use a custom dataset, refer to the dataset preparation guide [here](./assets/dataset.md).
 
+> [!NOTE]
+> To lower memory requirements:
+> - Pass `--precompute_conditions` when launching training.
+> - Do not perform validation/testing. This saves a significant amount of memory, which can be used to focus solely on training if you're on smaller VRAM GPUs.
```
sayakpaul, line 389: Maybe you have considered this already but gradient checkpointing and using a low-bit optimizer do help quite a bit.

finetrainers/args.py
```diff
@@ -234,6 +251,12 @@ def _add_model_arguments(parser: argparse.ArgumentParser) -> None:
         default=None,
         help="The directory where the downloaded models and datasets will be stored.",
     )
+    parser.add_argument("--text_encoder_dtype", type=str, default="bf16", help="Data type for the text encoder.")
+    parser.add_argument("--text_encoder_2_dtype", type=str, default="bf16", help="Data type for the text encoder 2.")
+    parser.add_argument("--text_encoder_3_dtype", type=str, default="bf16", help="Data type for the text encoder 3.")
+    parser.add_argument("--transformer_dtype", type=str, default="bf16", help="Data type for the transformer model.")
+    parser.add_argument("--unet_dtype", type=str, default="bf16", help="Data type for the U-Net model.")
```
sayakpaul, line 258: Do we have any `unet`-based model? :D

finetrainers/dataset.py
```diff
@@ -257,6 +261,32 @@ def _find_nearest_resolution(self, height, width):
...
+
+    def __len__(self) -> int:
+        return len(self.latent_conditions)
+
+    def __getitem__(self, index: int) -> Dict[str, Any]:
+        conditions = {}
+        latent_path = self.latents_path / self.latent_conditions[index]
+        condition_path = self.conditions_path / self.text_conditions[index]
+        conditions["latent_conditions"] = torch.load(latent_path, map_location="cpu", weights_only=True)
+        conditions["text_conditions"] = torch.load(condition_path, map_location="cpu", weights_only=True)
```
sayakpaul, line 286: Works for me!

finetrainers/hunyuan_video/hunyuan_video_lora.py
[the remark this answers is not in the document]
sayakpaul: Yeah I guess that is an extreme case. Folks that can afford to do validation during training, would probably have the GPU resources to run without precomputing the conditions for validation. If it becomes a problem, we can revisit but for now, without it is enough IMO.

finetrainers/ltx_video/ltx_video_lora.py
```diff
@@ -114,19 +127,55 @@ def prepare_latents(
...
+    image_or_video = image_or_video.to(device=device, dtype=vae.dtype)
     image_or_video = image_or_video.permute(0, 2, 1, 3, 4).contiguous()  # [B, C, F, H, W] -> [B, F, C, H, W]
-    latents = vae.encode(image_or_video).latent_dist.sample(generator=generator)
-    _, _, num_frames, height, width = latents.shape
-    latents = _normalize_latents(latents, vae.latents_mean, vae.latents_std)
+    if not precompute:
+        latents = vae.encode(image_or_video).latent_dist.sample(generator=generator)
+        latents = latents.to(dtype=dtype)
+        _, _, num_frames, height, width = latents.shape
+        latents = _normalize_latents(latents, vae.latents_mean, vae.latents_std)
```
sayakpaul, line 144:
```suggestion
        latents = _normalize_latents(latents, vae.config.latents_mean, vae.configlatents_std)
```

finetrainers/trainer.py
```diff
@@ -73,35 +80,93 @@ def __init__(self, args: Args) -> None:
...
         self._init_logging()
         self._init_directories_and_repositories()
 
         self.state.model_name = self.args.model_name
         self.model_config = get_config_from_model_name(self.args.model_name, self.args.training_type)
 
-    def prepare_models(self) -> None:
-        logger.info("Initializing models")
+    def prepare_dataset(self) -> None:
+        # TODO(aryan): Make a background process for fetching
```
sayakpaul, line 94: Is this `prefetch_factor` we're talking about? If not, maybe a more elaborate comment would be better.

finetrainers/trainer.py
```diff
@@ -692,7 +925,16 @@ def _init_directories_and_repositories(self) -> None:
...
+        if self.text_encoder is not None:
+            self.text_encoder = self.text_encoder.to(self.state.accelerator.device)
+        if self.text_encoder_2 is not None:
+            self.text_encoder_2 = self.text_encoder_2.to(self.state.accelerator.device)
+        if self.text_encoder_3 is not None:
+            self.text_encoder_3 = self.text_encoder_3.to(self.state.accelerator.device)
+        if self.transformer is not None:
+            self.transformer = self.transformer.to(self.state.accelerator.device)
+        if self.unet is not None:
+            self.unet = self.unet.to(self.state.accelerator.device)
```
sayakpaul, line 938: No biggie but we currently don't have any decent model that is based on unets, no? So, perhaps we can remove it altogether and support as in when needed?

sayakpaul (approved):
Just some minor nits. Otherwise, looks good.

README.md
a-r-r-o-w: Yep, should probably mentioned. We need gradient checkpointing for the numbers mentioned. Will update

finetrainers/args.py
a-r-r-o-w: I do have a local branch for training SD1.5 :) Helps test things faster

finetrainers/ltx_video/ltx_video_lora.py
a-r-r-o-w: It is not part of the config. All the LTX checkpoints have it in their state dict. We cannot apply this. Especially because we are considering only hosting the latest revision of the models as diffusers-format. For all other versions, we will use single file loading.

`latents_mean` and `latents_std` are registered buffers in the model implementation.

https://github.com/huggingface/diffusers/blob/ea1ba0ba53bdd6569547e26e518f094745ed9d03/src/diffusers/models/autoencoders/autoencoder_kl_ltx.py#L806

finetrainers/args.py
sayakpaul: Since the arg list already quite big maybe it won't hurt to try to remove arguments we don't want to support models we don't want to thoroughly test. WDYT? No strong opinions.

finetrainers/ltx_video/ltx_video_lora.py
sayakpaul: Oh :/

finetrainers/trainer.py
a-r-r-o-w: Currently, loading the dataset is a blocking process. We load one thing and process that thing and repeat. Prefetching here with any factor does not really help. We don't get the desired effect because the main process will still be slowed down due to loading and dispatching. The process of loading from disk to memory and converting to tensor, if done on a separate background thread is faster, and better overlaps for overlapping with computation from my preliminary tests.

finetrainers/trainer.py
a-r-r-o-w: We don't support here, but I've been exploring an ImageDataset and SD1.5 on the side in a separate branch. This is because:
- HunyuanVideo can learn styles solely from images without any video training
- SD1.5 is extremely fast to test things with

I think it's okay to have here for testing purposes

README.md
a-r-r-o-w: We don't support low-bit optimizers yet btw. Atleast the relevant arguments for that are not used. Would you like to open a PR adding support for TorchAO and prodigy optimizers?

finetrainers/args.py
a-r-r-o-w: Could you tell me which ones are not useful and can be removed?

README.md
sayakpaul: Just supporting 8bit AdamW from bitsandbytes should be fine as it is quite popular and have widely been tested. 

I can open a PR for that.

finetrainers/args.py
sayakpaul: `unet_dtype` as it’s ot relevant for the models we currently want to support.

a-r-r-o-w: Removed

README.md
a-r-r-o-w: CPUOffloadOptimizer from TorchAO too atleast, please. It is just too good to not support

sayakpaul: Sure I can do it one by one.

finetrainers/trainer.py
sayakpaul: Interesting. Definitely eager to see it.

a-r-r-o-w:
I believe I've addressed all reviews. Proceeding with merge. Follow-ups:
- BnB optimizers (@sayakpaul)
- TorchAO CPUOffloadOptimizer (@sayakpaul)
- Investigate why `batch_size` causes an error #132 (@a-r-r-o-w)
- CogVideoX T2V and I2V LoRA (@sayakpaul)
- Mochi LoRA (@a-r-r-o-w)
- FP8 layerwise upcasting training (@a-r-r-o-w)
- TorchAO quantization config training from diffusers (@a-r-r-o-w)
- LTX Full finetune (almost already ready and just needs long experimentation run) (@a-r-r-o-w)

sayakpaul:
Works for me, thanks for chalking that out. I can also support training with bnb-quantized checkpoints similar to https://github.com/huggingface/diffusers/tree/main/examples/research_projects/flux_lora_quantization (something that has worked like wonders in the LLM world and also for Flux) ;)

sayakpaul:
Will add:

- [ ] DeepSpeed support
- [ ] Separating trainer classes w.r.t. algorithms

PR #150, opened 2024-12-25
LTX Image2Video LoRA

a-r-r-o-w:
In order to run the following example, one needs atleast 24 GB VRAM (because of using 161 frames. if you set the resolution buckets to use lower amount of frames, vram requirements will be lower).

In order to run, one needs the same dataset format that we've been using so far:
- `prompts.txt`
- `videos.txt`
- `videos/`

Since this is `image-to-video` training, one also needs validation images in addition to prompts. As an example, this can be done by simply taking the first frame of your training videos:

```bash
mkdir -p /path/to/dataset/images
cd /path/to/dataset/videos

# For Windows, you will have to convert this bash snippet to Powershell/CMD
for file in *; do ffmpeg -y -i "$file" -frames:v 1 "../images/${file%.*}.png"; done
```

Ideally, you should test with different starting images other than your training videos to verify if the LoRA works.

<details>
<summary> script </summary>

```bash
#!/bin/bash

# export TORCH_LOGS="+dynamo,recompiles,graph_breaks"
# export TORCHDYNAMO_VERBOSE=1
export WANDB_MODE="offline"
export NCCL_P2P_DISABLE=1

GPU_IDS="2,3"

DATA_ROOT="/raid/aryan/video-dataset-disney"
CAPTION_COLUMN="prompts.txt"
VIDEO_COLUMN="videos.txt"

ID_TOKEN="afkx"

# Model arguments
model_cmd="--model_name ltx_video \
  --pretrained_model_name_or_path a-r-r-o-w/LTX-Video-diffusers"

# Dataset arguments
  --id_token $ID_TOKEN \
  --video_resolution_buckets 49x512x768 161x512x768 \
  --precompute_conditions"

# Dataloader arguments
dataloader_cmd="--dataloader_num_workers 0"

# Diffusion arguments
diffusion_cmd="--flow_resolution_shifting"

# Training arguments
training_cmd="--training_type lora_i2v \
  --seed 42 \
  --mixed_precision bf16 \
  --batch_size 1 \
  --train_steps 1000 \
  --gradient_checkpointing \
  --checkpointing_steps 500 \
  --checkpointing_limit 2 \
  --enable_slicing \
  --enable_tiling"

# Optimizer arguments
  --lr 1e-4 \
  --lr_scheduler cosine_with_restarts \
  --lr_warmup_steps 100 \
  --lr_num_cycles 1 \
  --beta1 0.9 \
  --beta2 0.95 \
  --weight_decay 1e-4 \
  --epsilon 1e-8 \
  --max_grad_norm 1.0"

# Validation arguments
validation_cmd="--validation_prompts \"$ID_TOKEN A black and white animated scene unfolds with an anthropomorphic goat surrounded by musical notes and symbols, suggesting a playful environment. Mickey Mouse appears, leaning forward in curiosity as the goat remains still. The goat then engages with Mickey, who bends down to converse or react. The dynamics shift as Mickey grabs the goat, potentially in surprise or playfulness, amidst a minimalistic background. The scene captures the evolving relationship between the two characters in a whimsical, animated setting, emphasizing their interactions and emotions.@@@49x512x768:::$ID_TOKEN A static black and white scene of three anthropomorphic characters is depicted in a series of animated frames. The first character, with oversized shoes, sings or speaks into a megaphone held by the central figure, who has an exaggerated open mouth. The third character, wearing a chef's hat, controls the megaphone's volume. Musical notes surround the characters, hinting at music, but there is no movement. The plain background focuses attention on the characters' interaction, maintaining a consistent composition throughout the scene.@@@49x512x768:::$ID_TOKEN A black and white animated scene unfolds with an anthropomorphic goat surrounded by musical notes and symbols, suggesting a playful environment. Mickey Mouse appears, leaning forward in curiosity as the goat remains still. The goat then engages with Mickey, who bends down to converse or react. The dynamics shift as Mickey grabs the goat, potentially in surprise or playfulness, amidst a minimalistic background. The scene captures the evolving relationship between the two characters in a whimsical, animated setting, emphasizing their interactions and emotions.@@@161x512x768:::$ID_TOKEN A static black and white scene of three anthropomorphic characters is depicted in a series of animated frames. The first character, with oversized shoes, sings or speaks into a megaphone held by the central figure, who has an exaggerated open mouth. The third character, wearing a chef's hat, controls the megaphone's volume. Musical notes surround the characters, hinting at music, but there is no movement. The plain background focuses attention on the characters' interaction, maintaining a consistent composition throughout the scene.@@@161x512x768:::$ID_TOKEN Three mechanical towers, each with unique designs on their tops, are connected by pipes on a flat surface against a cloudy black-and-white sky. Rain clouds appear, and water droplets fall, setting an industrial scene. The scene shifts to a ship's interior, where two cartoon characters interact amidst nautical elements. A claw-like appendage emerges, reaching for the ship, causing tension. The appendage extends, tilting the ship dramatically, introducing movement and a sense of urgency, highlighting potential danger within the ship's confines.@@@161x512x768:::$ID_TOKEN A tranquil, monochromatic scene unfolds with an animated character standing on a dock beside a body of water, a lighthouse visible in the background. The character is startled, running away from the viewer's perspective. A mechanical device with levers and a counter is introduced, attached to a rope, with a smaller, humanoid figure near it. The smaller figure engages with the machine, pulling the rope, while the larger figure remains in motion, creating a sense of tension and urgency in this classic, hand-drawn animated sequence.@@@161x512x768:::$ID_TOKEN A dynamic sequence unfolds on the deck of a ship, where a small, mouse-like character with large ears and short pants enthusiastically steers the vessel using a wheel. A larger, bulky character with a long pole engages in a playful confrontation, asserting dominance or playfully provoking the smaller one. Expressive gestures and movements convey emotions and intentions, set against a nautical backdrop featuring a steering wheel, life preserver, and bell. The two characters interact in a lively, competitive, or friendly exchange.@@@161x512x768:::$ID_TOKEN Mickey Mouse, with his distinctive round ears and black gloves, crouches attentively beside a pig character lying on the ground. The pig undergoes subtle movements, possibly preparing for action, while Mickey's posture and expression change as they interact. Mickey's mouth opens in surprise, and his body language suggests engagement. The pig raises its head or moves its limbs, indicating alertness or response. The background remains unchanged, with consistent lighting and shading highlighting the characters against a plain backdrop, capturing a series of moments of communication and reaction between the two.@@@161x512x768\" \
  --validation_images \"$DATA_ROOT/images/a3c275fc2eb0a67168a7c58a6a9adb14.png:::$DATA_ROOT/images/bf06573576ae0ea4d27a178b4d6e95a1.png:::$DATA_ROOT/images/a3c275fc2eb0a67168a7c58a6a9adb14.png:::$DATA_ROOT/images/bf06573576ae0ea4d27a178b4d6e95a1.png:::$DATA_ROOT/images/1094139d474e65852826d64a1b4aa520.png:::$DATA_ROOT/images/3108dd567bd8669967bc83e0bc50dab2.png:::$DATA_ROOT/images/1d50a3d9703f152758d5422c8b48010f.png:::$DATA_ROOT/images/12e51adf1acbf7acbb703a96a464a39b.png\" \
  --num_validation_videos 1 \
  --validation_steps 50"


REGULARIZATION_CMDS=("--caption_dropout_p 0.00" "--caption_dropout_p 0.05" "--caption_dropout_p 0.05 --image_condition_dropout_p 0.05 --image_condition_noise_scale 0.00" "--caption_dropout_p 0.05 --image_condition_dropout_p 0.05 --image_condition_noise_scale 0.1" "--caption_dropout_p 0.05 --image_condition_dropout_p 0.1 --image_condition_noise_scale 0.1")

LORA_CONFIGS=("--rank 128 --lora_alpha 64" "--rank 128 --lora_alpha 128" "--rank 128 --lora_alpha 256" "--rank 64 --lora_alpha 64" "--rank 256 --lora_alpha 256")

for regularization_cmd in "${REGULARIZATION_CMDS[@]}"
do
  sanitized_regularization_cmd=$(echo "$regularization_cmd" | sed 's/--//g; s/ /_/g; s/=/-/g')
  
  miscellaneous_cmd="--tracker_name finetrainers-ltxv \
    --output_dir /raid/aryan/ltx-video-$sanitized_regularization_cmd \
    --nccl_timeout 1800 \
    --report_to wandb"
  
  cmd="accelerate launch --config_file accelerate_configs/uncompiled_2.yaml --gpu_ids $GPU_IDS train.py \
    $model_cmd \
    $dataset_cmd \
    $dataloader_cmd \
    $diffusion_cmd \
    $training_cmd --rank 128 --lora_alpha 128 \
    $optimizer_cmd \
    $validation_cmd \
    $regularization_cmd \
    $lora_config \
    $miscellaneous_cmd"

  echo "Running command: $cmd"
  eval $cmd
done

for lora_config in "${LORA_CONFIGS[@]}"
do
  sanitized_lora_config=$(echo "$lora_config" | sed 's/--//g; s/ /_/g; s/=/-/g')
  regularization_cmd="--caption_dropout_p 0.05"
  
  miscellaneous_cmd="--tracker_name finetrainers-ltxv \
    --output_dir /raid/aryan/ltx-video-$sanitized_lora_config \
    --nccl_timeout 1800 \
    --report_to wandb"
  
  cmd="accelerate launch --config_file accelerate_configs/uncompiled_2.yaml --gpu_ids $GPU_IDS train.py \
    $model_cmd \
    $dataset_cmd \
    $dataloader_cmd \
    $diffusion_cmd \
    $training_cmd \
    $optimizer_cmd \
    $validation_cmd \
    $regularization_cmd \
    $lora_config \
    $miscellaneous_cmd"

  echo "Running command: $cmd"
  eval $cmd
done

```
</details>

Note that I don't know if the training works yet. I've queued some runs that should finish overnight if there were no bugs that would cause a crash.

### DIFF
--- a/finetrainers/args.py
+++ b/finetrainers/args.py
@@ -11,6 +11,7 @@ class Args:
     The arguments for the finetrainers training script.
 
     Args:
+        TODO: write informational docstring
         flow_resolution_shifting (`bool`, defaults to `False`):
             Resolution-dependant shifting of timestep schedules.
             [Scaling Rectified Flow Transformers for High-Resolution Image Synthesis](https://arxiv.org/abs/2403.03206)
@@ -37,8 +38,6 @@ class Args:
     image_resolution_buckets: List[Tuple[int, int]] = None
     video_resolution_buckets: List[Tuple[int, int, int]] = None
     video_reshape_mode: Optional[str] = None
-    caption_dropout_p: float = 0.00
-    caption_dropout_technique: str = "empty"
     precompute_conditions: bool = False
 
     # Dataloader arguments
@@ -56,6 +55,15 @@ class Args:
     flow_logit_mean: float = 0.0
     flow_logit_std: float = 1.0
     flow_mode_scale: float = 1.29
+
+    # Regularization arguments
+    caption_dropout_p: float = 0.00
+    caption_dropout_technique: str = "empty"
+    image_condition_dropout_p: float = 0.00
+    image_condition_noise_scale: float = 0.00
+    image_condition_noise_type: float = "gaussian"
+    # We are currently only supporting adding noise post-encoding for regularization. For models like CogVideoX
+    # noise is added to image before the vae encoding. TODO: add support for this.
 
     # Training arguments
     training_type: str = None
@@ -136,13 +144,18 @@ class Args:
                 "image_resolution_buckets": self.image_resolution_buckets,
                 "video_resolution_buckets": self.video_resolution_buckets,
                 "video_reshape_mode": self.video_reshape_mode,
-                "caption_dropout_p": self.caption_dropout_p,
-                "caption_dropout_technique": self.caption_dropout_technique,
                 "precompute_conditions": self.precompute_conditions,
             },
             "dataloader_arguments": {
                 "dataloader_num_workers": self.dataloader_num_workers,
                 "pin_memory": self.pin_memory,
+            },
+            "regularization_arguments": {
+                "caption_dropout_p": self.caption_dropout_p,
+                "caption_dropout_technique": self.caption_dropout_technique,
+                "image_condition_dropout_p": self.image_condition_dropout_p,
+                "image_condition_noise_scale": self.image_condition_noise_scale,
+                "image_condition_noise_type": self.image_condition_noise_type,
             },
             "training_arguments": {
                 "training_type": self.training_type,
@@ -207,6 +220,7 @@ def parse_arguments() -> Args:
     _add_dataset_arguments(parser)
     _add_dataloader_arguments(parser)
     _add_diffusion_arguments(parser)
+    _add_regularization_arguments(parser)
     _add_training_arguments(parser)
     _add_optimizer_arguments(parser)
     _add_validation_arguments(parser)
@@ -325,19 +339,6 @@ def _add_dataset_arguments(parser: argparse.ArgumentParser) -> None:
         help="All input videos are reshaped to this mode. Choose between ['center', 'random', 'none']",
     )
     parser.add_argument(
-        "--caption_dropout_p",
-        type=float,
-        default=0.00,
-        help="Probability of dropout for the caption tokens.",
-    )
-    parser.add_argument(
-        "--caption_dropout_technique",
-        type=str,
-        default="empty",
-        choices=["empty", "zero"],
-        help="Technique to use for caption dropout.",
-    )
-    parser.add_argument(
         "--precompute_conditions",
         action="store_true",
         help="Whether or not to precompute the conditionings for the model.",
@@ -388,6 +389,41 @@ def _add_diffusion_arguments(parser: argparse.ArgumentParser) -> None:
         type=float,
         default=1.29,
         help="Scale of mode weighting scheme. Only effective when using the `'mode'` as the `weighting_scheme`.",
+    )
+
+
+def _add_regularization_arguments(parser: argparse.ArgumentParser) -> None:
+    parser.add_argument(
+        "--caption_dropout_p",
+        type=float,
+        default=0.00,
+        help="Probability of dropout for the caption tokens.",
+    )
+    parser.add_argument(
+        "--caption_dropout_technique",
+        type=str,
+        default="empty",
+        choices=["empty", "zero"],
+        help="Technique to use for caption dropout.",
+    )
+    parser.add_argument(
+        "--image_condition_dropout_p",
+        type=float,
+        default=0.00,
+        help="Probability of dropout for the image condition latents. This will replace the image latent condition with a zero tensor.",
+    )
+    parser.add_argument(
+        "--image_condition_noise_scale",
+        type=float,
+        default=0.00,
+        help="Maximum scale of noise to add to the image condition latents. A random noise tensor will be added to the image condition latents at at strength between 0 and this value.",
+    )
+    parser.add_argument(
+        "--image_condition_noise_type",
+        type=str,
+        default="gaussian",
+        choices=["gaussian"],
+        help="Type of noise to add to the image condition latents. Defaults to plain Gaussian noise.",
     )
 
 
@@ -704,8 +740,6 @@ def _map_to_args_type(args: Dict[str, Any]) -> Args:
     result_args.image_resolution_buckets = args.image_resolution_buckets or DEFAULT_IMAGE_RESOLUTION_BUCKETS
     result_args.video_resolution_buckets = args.video_resolution_buckets or DEFAULT_VIDEO_RESOLUTION_BUCKETS
     result_args.video_reshape_mode = args.video_reshape_mode
-    result_args.caption_dropout_p = args.caption_dropout_p
-    result_args.caption_dropout_technique = args.caption_dropout_technique
     result_args.precompute_conditions = args.precompute_conditions
 
     # Dataloader arguments
@@ -718,6 +752,13 @@ def _map_to_args_type(args: Dict[str, Any]) -> Args:
     result_args.flow_logit_mean = args.flow_logit_mean
     result_args.flow_logit_std = args.flow_logit_std
     result_args.flow_mode_scale = args.flow_mode_scale
+
+    # Regularization arguments
+    result_args.caption_dropout_p = args.caption_dropout_p
+    result_args.caption_dropout_technique = args.caption_dropout_technique
+    result_args.image_condition_dropout_p = args.image_condition_dropout_p
+    result_args.image_condition_noise_scale = args.image_condition_noise_scale
+    result_args.image_condition_noise_type = args.image_condition_noise_type
 
     # Training arguments
     result_args.training_type = args.training_type
--- a/finetrainers/ltx_video/ltx_video_lora.py
+++ b/finetrainers/ltx_video/ltx_video_lora.py
@@ -1,9 +1,16 @@
+import random
 from typing import Dict, List, Optional, Union
 
 import torch
 import torch.nn as nn
 from accelerate.logging import get_logger
-from diffusers import AutoencoderKLLTXVideo, FlowMatchEulerDiscreteScheduler, LTXPipeline, LTXVideoTransformer3DModel
+from diffusers import (
+    AutoencoderKLLTXVideo,
+    FlowMatchEulerDiscreteScheduler,
+    LTXPipeline,
+    LTXImageToVideoPipeline,
+    LTXVideoTransformer3DModel,
+)
 from diffusers.utils import logging
 from transformers import T5EncoderModel, T5Tokenizer
 from PIL import Image
@@ -53,7 +60,7 @@ def load_diffusion_models(
     return {"transformer": transformer, "scheduler": scheduler}
 
 
-def initialize_pipeline(
+def initialize_pipeline_t2v(
     model_id: str = "Lightricks/LTX-Video",
     text_encoder_dtype: torch.dtype = torch.bfloat16,
     transformer_dtype: torch.dtype = torch.bfloat16,
@@ -101,6 +108,54 @@ def initialize_pipeline(
     return pipe
 
 
+def initialize_pipeline_i2v(
+    model_id: str = "Lightricks/LTX-Video",
+    text_encoder_dtype: torch.dtype = torch.bfloat16,
+    transformer_dtype: torch.dtype = torch.bfloat16,
+    vae_dtype: torch.dtype = torch.bfloat16,
+    tokenizer: Optional[T5Tokenizer] = None,
+    text_encoder: Optional[T5EncoderModel] = None,
+    transformer: Optional[LTXVideoTransformer3DModel] = None,
+    vae: Optional[AutoencoderKLLTXVideo] = None,
+    scheduler: Optional[FlowMatchEulerDiscreteScheduler] = None,
+    device: Optional[torch.device] = None,
+    revision: Optional[str] = None,
+    cache_dir: Optional[str] = None,
+    enable_slicing: bool = False,
+    enable_tiling: bool = False,
+    enable_model_cpu_offload: bool = False,
+    **kwargs,
+) -> LTXImageToVideoPipeline:
+    component_name_pairs = [
+        ("tokenizer", tokenizer),
+        ("text_encoder", text_encoder),
+        ("transformer", transformer),
+        ("vae", vae),
+        ("scheduler", scheduler),
+    ]
+    components = {}
+    for name, component in component_name_pairs:
+        if component is not None:
+            components[name] = component
+
+    pipe = LTXImageToVideoPipeline.from_pretrained(model_id, **components, revision=revision, cache_dir=cache_dir)
+    pipe.text_encoder = pipe.text_encoder.to(dtype=text_encoder_dtype)
+    pipe.transformer = pipe.transformer.to(dtype=transformer_dtype)
+    pipe.vae = pipe.vae.to(dtype=vae_dtype)
+
+    if enable_slicing:
+        pipe.vae.enable_slicing()
+    if enable_tiling:
+        pipe.vae.enable_tiling()
+
+    if enable_model_cpu_offload:
+        pipe.enable_model_cpu_offload(device=device)
+    else:
+        pipe.to(device=device)
+
+    return pipe
+
+
 def prepare_conditions(
     tokenizer: T5Tokenizer,
     text_encoder: T5EncoderModel,
@@ -119,7 +174,7 @@ def prepare_conditions(
     return _encode_prompt_t5(tokenizer, text_encoder, prompt, device, dtype, max_sequence_length)
 
 
-def prepare_latents(
+def prepare_latents_t2v(
     vae: AutoencoderKLLTXVideo,
     image_or_video: torch.Tensor,
     patch_size: int = 1,
@@ -150,6 +205,7 @@ def prepare_latents(
             h = torch.cat(encoded_slices)
         else:
             h = vae._encode(image_or_video)
+
         _, _, num_frames, height, width = h.shape
 
         # TODO(aryan): This is very stupid that we might possibly be storing the latents_mean and latents_std in every file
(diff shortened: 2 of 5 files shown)


finetrainers/trainer.py
```diff
@@ -851,7 +867,18 @@ def validate(self, step: int) -> None:
...
+                    tracker.log(
+                        {
+                            "validation": {"images": image_artifacts, "videos": video_artifacts},
+                        },
+                        step=step,
+                    )
+
+        # Remove all hooks that might have been added during pipeline initialization to the models
+        pipeline.remove_all_hooks()
+        del pipeline
```
a-r-r-o-w, line 881:
```suggestion
```

finetrainers/args.py
```diff
@@ -11,6 +11,7 @@ class Args:
     The arguments for the finetrainers training script.
 
     Args:
+        TODO: write informational docstring
```
sayakpaul, line 14: Yeah this should be clubbed in separate PRs. Just TODO note is fine!

finetrainers/args.py
```diff
@@ -391,6 +392,41 @@ def _add_diffusion_arguments(parser: argparse.ArgumentParser) -> None:
     )
 
 
+def _add_regularization_arguments(parser: argparse.ArgumentParser) -> None:
```
sayakpaul, line 395: Beautiful!

finetrainers/args.py
```diff
@@ -391,6 +392,41 @@ def _add_diffusion_arguments(parser: argparse.ArgumentParser) -> None:
...
+    parser.add_argument(
+        "--caption_dropout_p",
```
sayakpaul, line 397: Let's validate this to always be in [0, 1] if not already?

finetrainers/args.py
```diff
@@ -391,6 +392,41 @@ def _add_diffusion_arguments(parser: argparse.ArgumentParser) -> None:
...
+    )
+    parser.add_argument(
+        "--caption_dropout_technique",
+        type=str,
+        default="empty",
+        choices=["empty", "zero"],
+        help="Technique to use for caption dropout.",
+    )
+    parser.add_argument(
+        "--image_condition_dropout_p",
```
sayakpaul, line 410: Same as above.

finetrainers/args.py
```diff
@@ -391,6 +392,41 @@ def _add_diffusion_arguments(parser: argparse.ArgumentParser) -> None:
...
+        default=0.00,
+        help="Maximum scale of noise to add to the image condition latents. A random noise tensor will be added to the image condition latents at at strength between 0 and this value.",
+    )
+    parser.add_argument(
+        "--image_condition_noise_type",
+        type=str,
+        default="gaussian",
+        choices=["gaussian"],
+        help="Type of noise to add to the image condition latents. Defaults to plain Gaussian noise.",
+    )
```
sayakpaul, line 427: Does this need to be different for Cog I2V?

finetrainers/args.py
```diff
@@ -391,6 +392,41 @@ def _add_diffusion_arguments(parser: argparse.ArgumentParser) -> None:
...
+        help="Technique to use for caption dropout.",
+    )
+    parser.add_argument(
+        "--image_condition_dropout_p",
+        type=float,
+        default=0.00,
+        help="Probability of dropout for the image condition latents. This will replace the image latent condition with a zero tensor.",
+    )
+    parser.add_argument(
+        "--image_condition_noise_scale",
```
sayakpaul, line 416: Same as above.

finetrainers/ltx_video/ltx_video_lora.py
```diff
@@ -165,7 +221,90 @@ def prepare_latents(
...
+
+    if image_or_video.size(2) == 1:
+        logger.warning("Received a single frame for LTX Video I2V training. Duplicating the frame to create a video.")
+        # Duplicate the image to create atleast one frame to run prediction on
+        image_or_video = image_or_video.repeat(1, 1, 2, 1, 1)
+
+    image_or_video = image_or_video.to(device=device, dtype=vae.dtype)
+    image_or_video = image_or_video.permute(0, 2, 1, 3, 4).contiguous()  # [B, C, F, H, W] -> [B, F, C, H, W]
+
+    # Note: we separately encode the image and video because there is a 4x compression applied. We only want to condition
```
sayakpaul, line 248: For my underatanding:

If we encode `image_or_video` without this indexing and then index the `latents` post-encoding, what's the difference? Because of the compression factor?

finetrainers/ltx_video/ltx_video_lora.py
```diff
@@ -165,7 +221,90 @@ def prepare_latents(
...
+            image, video = image_or_video[:, :, :1], image_or_video[:, :, 1:]
+            h_image = vae._encode(image)
+            h_video = vae._encode(video)
+            h = torch.cat([h_image, h_video], dim=2)
+
+        _, _, num_frames, height, width = h.shape
+
+        # TODO(aryan): This is very stupid that we might possibly be storing the latents_mean and latents_std in every file
+        # if precomputation is enabled. We should probably have a single file where re-usable properties like this are stored
+        # so as to reduce the disk memory requirements of the precomputed files.
```
sayakpaul, line 281: Anything to make it simple for now is welcome. It's a very tiny tensor so, very much okay with this

finetrainers/ltx_video/ltx_video_lora.py
```diff
@@ -174,8 +313,30 @@ def post_latent_preparation(
...
+    # We know that the first frame is the image condition frame
+    if image_condition_dropout_p > 0.0:
+        # TODO(aryan): The zero-ing dropout is for testing based on ideas from CogVideoX. Since there is no source to
+        # refer to for LTX Video, this is experimental and should be unsupported if it does not work.
+        if random.random() < image_condition_dropout_p:
+            latents[:, :, 0].zero_()
+        else:
+            # Map from [0, 1] to [0, image_condition_noise_scale]
+            scale_factor = random.random() * image_condition_noise_scale
+            # :/ Because we don't have torch.randn_like
```
sayakpaul, line 335: Are we referring to [this](https://pytorch.org/docs/stable/generated/torch.randn_like.html)? Also, do we always have to add noise to the conditional latent? For Flux Control, we keep the conditional latent clean.

finetrainers/ltx_video/ltx_video_lora.py
```diff
@@ -165,7 +221,90 @@ def prepare_latents(
...
+
+    image_or_video = image_or_video.to(device=device, dtype=vae.dtype)
+    image_or_video = image_or_video.permute(0, 2, 1, 3, 4).contiguous()  # [B, C, F, H, W] -> [B, F, C, H, W]
+
+    # Note: we separately encode the image and video because there is a 4x compression applied. We only want to condition
+    # on the first frame of the video, and not the first 4 frames.
+    image, video = image_or_video[:, :, :1], image_or_video[:, :, 1:]
+
+    if not precompute:
+        image_latents = vae.encode(image).latent_dist.sample(generator=generator)
```
sayakpaul, line 253: Not important for this PR but just sharing info. 

Some works like InstructPix2Pix compute a mode, instead on the condition.

finetrainers/ltx_video/ltx_video_lora.py
```diff
@@ -187,6 +348,32 @@ def collate_fn_t2v(batch: List[List[Dict[str, torch.Tensor]]]) -> Dict[str, torc
...
+    noise: torch.Tensor,
+    latents: torch.Tensor,
+    num_frames: int,
+    height: int,
+    width: int,
+    **kwargs,
+) -> torch.Tensor:
+    # We do not add noise to the first frame as it is what we want to condition on.
+    image_frame_end_offset = 1 * height * width
+    image_latents = latents[:, :image_frame_end_offset]
```
sayakpaul, line 369: Can't we do `image_latents = latents[:, :, 0]`? Or is it more future-proof?

finetrainers/ltx_video/ltx_video_lora.py
```diff
@@ -187,6 +348,32 @@ def collate_fn_t2v(batch: List[List[Dict[str, torch.Tensor]]]) -> Dict[str, torc
...
+    video_noisy_latents = (1.0 - sigmas) * latents[:, image_frame_end_offset:] + sigmas * noise[
+        :, image_frame_end_offset:
+    ]
+    noisy_latents = torch.cat([image_latents, video_noisy_latents], dim=1)
```
sayakpaul, line 373: Shouldn't this be `noisy_latents = torch.cat([video_noisy_latents, image_latents], dim=1)`?

This is what we do in Cog, too:

https://github.com/a-r-r-o-w/finetrainers/blob/2a594ea2dbea5e9701c1c3494524a567f3071f83/training/cogvideox/cogvideox_image_to_video_lora.py#L806

finetrainers/trainer.py
```diff
@@ -646,7 +652,14 @@ def train(self) -> None:
...
                     )
-                    noisy_latents = (1.0 - sigmas) * latent_conditions["latents"] + sigmas * noise
+                    # TODO(aryan): It seems like we need to specialize for how to add noise to latents as well
+                    # This might have to be implemented for every model
+                    if "prepare_noisy_latents" in self.model_config.keys():
+                        noisy_latents = self.model_config["prepare_noisy_latents"](
+                            sigmas=sigmas, noise=noise, **latent_conditions
+                        )
+                    else:
+                        noisy_latents = (1.0 - sigmas) * latent_conditions["latents"] + sigmas * noise
```
sayakpaul, line 662: 👍

sayakpaul (commented):
Thanks!

finetrainers/args.py
a-r-r-o-w: I added this for testing purposes. I don't think I've been able to see its effectiveness with LTX yet so I will actually remove it.

Cog uses pre-encoding noise, while here I'm using post-encoding noise addition. This is inspired by what the LTX folks do in the original repository for finetuning

finetrainers/args.py
a-r-r-o-w: Will be removing this

finetrainers/ltx_video/ltx_video_lora.py
a-r-r-o-w: If we don't do the separation like done here, due to the time compression, the first latent frame will hold information from 4 frames of the original video instead of 1 frame. I've never seen this to be the case with many of the video models available today, so I made an assumption that this is how LTX is probably trained too (because we do the exact same thing for preparing latents during inference)

finetrainers/ltx_video/ltx_video_lora.py
a-r-r-o-w: Yes, torch.randn_like does not support using a `generator`, that's why I first create an empty tensor and then call `normal_` to create the gaussian noise.

Here, I tried making an educated guess that adding noise to the image would serve as a good regularizer because we don't have information on how it was trained. I haven't done enough experiments to see that it yields any benefits so I will be removing this

a-r-r-o-w: For Flux as well, we _chose_ to not add any noise, but typically this should be experimented with so that the model can learn to pay better attention to control signals even if they are noisy. I have seen this being done atleast a few places now, so assumed it would make sense but unless I run a large 10k-50k step run, it would be hard to evaluate its effect

finetrainers/ltx_video/ltx_video_lora.py
sayakpaul: Wow, that is cool. Thanks for explaining. Then let’s make this a note?

finetrainers/ltx_video/ltx_video_lora.py
sayakpaul: Makes sense.

finetrainers/ltx_video/ltx_video_lora.py
yoavhacohen: Since this is a causal VAE, the first latent frame always corresponds to the first pixel frame.
Other latent frames correspond to 8 pixels frames (not 4).
There's no need to encode the first frame separately.

sayakpaul: Ah the temporal compression factor [is](https://github.com/huggingface/diffusers/blob/1b202c5730631417000585e3639539cefc79cbd7/src/diffusers/pipelines/ltx/pipeline_ltx_image2video.py#L209) 8.

finetrainers/ltx_video/ltx_video_lora.py
sayakpaul: So, IMO, it's okay to keep the argument for experimentation purposes.

finetrainers/ltx_video/ltx_video_lora.py
sayakpaul: This and [this](https://github.com/a-r-r-o-w/finetrainers/pull/150/#discussion_r1897880866) -- suggests we should definitely experiment if noising the conditional latent is a good regularizer in this case.

a-r-r-o-w: @yoavhacohen I made an assumption that the image latent is always kept clean and no denoising is applied to it. Is that the case when training LTXV? Or do you add a varying amount of noise based on some randomly sampled timestep (different from that of other frame latents) and perform denoising on it too? This is the first time I've seen this technique of per-token/per-frame denoising level so I'm not sure what to do without making guesses :/

finetrainers/ltx_video/ltx_video_lora.py
```diff
@@ -174,8 +313,30 @@ def post_latent_preparation(
...
+    if image_condition_dropout_p > 0.0:
+        # TODO(aryan): The zero-ing dropout is for testing based on ideas from CogVideoX. Since there is no source to
+        # refer to for LTX Video, this is experimental and should be unsupported if it does not work.
+        if random.random() < image_condition_dropout_p:
+            latents[:, :, 0].zero_()
+        else:
+            # Map from [0, 1] to [0, image_condition_noise_scale]
+            scale_factor = random.random() * image_condition_noise_scale
+            # :/ Because we don't have torch.randn_like
+            latents[:, :, 0] = (
```
a-r-r-o-w, line 336: @yoavhacohen Would love to know your thoughts on this as well. It was an educated guess based on some other training code I've come across for image latent regularisation. Since we're not entirely sure how LTXV was trained, this may probably not be helpful and cause worse results. I'm still experimenting but any details of training for things like this would be super awesome 🤗

finetrainers/ltx_video/ltx_video_lora.py
a-r-r-o-w: Thanks @yoavhacohen, I'll make the update soon! Makes sense looking at it in hindsight, since this was mostly just guesswork

finetrainers/ltx_video/ltx_video_lora.py
yoavhacohen: I recommend training text-to-video and image-to-video models simultaneously, adding a varying amount of noise based on a randomly sampled small timestep.
For training with image conditioning, you just need to determine the noise scheduler to apply to the tokens that correspond to the conditioning frame (versus the other tokens).
I recommend making the implementation generic rather than specific to the first frame - it should support conditioning on any subset of tokens, not just those corresponding to the first frame.

finetrainers/ltx_video/ltx_video_lora.py
yoavhacohen: You should add noise to the conditioning tokens in the same way as the other tokens - just use a different noise level.

finetrainers/ltx_video/ltx_video_lora.py
sayakpaul: Thank you! And generally speaking, the noise to be added to the conditioning frame — should it have a smaller magnitude than the one being added to the rest of the tokens?

a-r-r-o-w: @yoavhacohen That sounds great, thank you! Since we don't support the idea of per-token timesteps in diffusers, I think we might have to write a custom scheduler step implementation - will give this a stab soon

yoavhacohen: The noise applied to conditioning tokens should be reduced compared to other tokens in the sequence - that's actually what defines them as conditioning tokens. You can think of this as a kind of soft mask.​​​​​​​​​​​​​​​​

UsernamesAreSilly4:
What are some settings I can use if I want to train on an RTX 3060 12 GB VRAM? I read that if you turn off validation, you only need 11 GB VRAM. What other optimizations can I use?

sayakpaul:
You could try `gradient_checkpointing`, `precompute_conditions`, `use_8bit_bnb`, `deespeed`.

scarbain:
Can we test this PR or the comments about the TODO changes are breaking training and not worth testing until done ? :)

a-r-r-o-w:
Hi @scarbain. Sorry, I haven't had the time to move this to completion yet. My last training run did not yield particularly interesting results, and I'm yet to address some comments from the original model author, so would recommend waiting until merged unless you have access to ample GPU resources for testing/debugging.

Will try and complete soon after a new suite of memory optimizations in the coming days

scarbain:
> Hi @scarbain. Sorry, I haven't had the time to move this to completion yet. My last training run did not yield particularly interesting results, and I'm yet to address some comments from the original model author, so would recommend waiting until merged unless you have access to ample GPU resources for testing/debugging.
> 
> Will try and complete soon after a new suite of memory optimizations in the coming days

Hi @a-r-r-o-w ! I'm sorry for asking again, I certainly don't want to pressure you on this because all your work on this repository is a great gift to the community and you should prioritise how you want. Do you have an approximate ETA for completing this PR ? :)

ArEnSc:
@a-r-r-o-w hey I am thinking about taking over this PR
what's left to do? maybe I can take over

a-r-r-o-w:
@ArEnSc See #245, I've added LTX condition based on my understanding of the paper. That PR is also adding a few more things and testing it is taking longer than expected :/

scarbain:
@a-r-r-o-w thanks for the update! What's your testing process ? I need to train a LoRA for LTX image condition so if I can help with testing in the same time :)

PR #158, opened 2024-12-27
Allow images; Remove LLM generated prefixes; Allow JSON/JSONL; Fix bugs

a-r-r-o-w:


finetrainers/dataset.py
```diff
@@ -45,28 +51,45 @@ def __init__(
         self.id_token = f"{id_token.strip()} " if id_token else ""
         self.resolution_buckets = resolution_buckets
 
-        # Two methods of loading data are supported.
+        # Three methods of loading data are supported.
         #   - Using a CSV: caption_column and video_column must be some column in the CSV. One could
         #     make use of other columns too, such as a motion score or aesthetic score, by modifying the
         #     logic in CSV processing.
         #   - Using two files containing line-separate captions and relative paths to videos.
+        #   - Using a JSON file containing a list of dictionaries, where each dictionary has a `caption_column` and `video_column` key.
```
sayakpaul, line 59: Like this?

```py
[{"prompt": ..., "video": ...}, ...]
```

finetrainers/dataset.py
```diff
@@ -45,28 +51,45 @@ def __init__(
...
             raise ValueError(
                 f"Expected length of prompts and videos to be the same but found {len(self.prompts)=} and {len(self.video_paths)=}. Please ensure that the number of caption prompts and videos match in your dataset."
             )
 
+        # Clean LLM start phrases
+        for i in range(len(self.prompts)):
+            self.prompts[i] = self.prompts[i].strip()
+            for phrase in COMMON_LLM_START_PHRASES:
+                if self.prompts[i].startswith(phrase):
+                    self.prompts[i] = self.prompts[i].removeprefix(phrase).strip()
```
sayakpaul, line 91: Should this be user-configured? 

And maybe we should also note this from our data-prep guide?

sayakpaul (approved):
Thanks!

finetrainers/trainer.py
```diff
@@ -751,13 +752,13 @@ def train(self) -> None:
...
-                )
-                if should_run_validation:
-                    self.validate(global_step)
+                    # Maybe run validation
+                    should_run_validation = (
+                        self.args.validation_every_n_steps is not None
+                        and global_step % self.args.validation_every_n_steps == 0
+                    )
+                    if should_run_validation:
+                        self.validate(global_step)
```
a-r-r-o-w, line 761: global_step is only incremented inside the `if accelerator.sync_gradients` statement. If we perform validation with `gradient_accumulation_steps > 1`, validation will run `gradient_accumulation_steps` many times until the next gradient sync happens. I believe this is correct approach

finetrainers/dataset.py
```diff
@@ -148,12 +189,47 @@ def _load_dataset_from_csv(self) -> Tuple[List[str], List[str]]:
...
+            )
+
+        return prompts, video_paths
+
+    def _preprocess_image(self, path: Path) -> torch.Tensor:
+        # TODO(aryan): Support alpha channel in future by whitening background
+        image = TTF.Image.open(path.as_posix()).convert("RGB")
+        image = TTF.to_tensor(image)
+        image = self.video_transforms(image)
+        image = image.unsqueeze(0).contiguous()  # [C, H, W] -> [C, 1, H, W] (1-frame video)
```
a-r-r-o-w, line 225:
```suggestion
        image = image.unsqueeze(0).contiguous()  # [C, H, W] -> [C, 1, H, W] (1-frame video)
```
```suggestion
        image = image.unsqueeze(0).contiguous()  # [C, H, W] -> [1, C, H, W] (1-frame video)
```

a-r-r-o-w:
After the HunyuanVideo fixes, the precomputed vs non-precomputed runs match almost exactly when starting with the same parameters. The weights converge to very similar values and the validation videos demonstrate this as well: https://api.wandb.ai/links/aryanvs/3aixk4xk

sayakpaul:
Is this ready for another review?

a-r-r-o-w:
No.

finetrainers/dataset.py
a-r-r-o-w: Yep. An example dataset would be this: https://huggingface.co/datasets/omni-research/DREAM-1K

a-r-r-o-w:
Since this makes some changes to the README, would prefer to merge after #175. Will move the dataset.md file into the `docs/` folder too here, because currently it's in `assets/`

sayakpaul:
SG!

LMK if you would like me to review, too.

finetrainers/trainer.py
```diff
@@ -955,7 +956,9 @@ def validate(self, step: int, final_validation: bool = False) -> None:
                 width=width,
                 num_frames=num_frames,
                 num_videos_per_prompt=self.args.num_validation_videos_per_prompt,
-                generator=self.state.generator,
+                generator=torch.Generator(device=accelerator.device).manual_seed(
```
a-r-r-o-w, line 959: Here, if we use the state, we get different validation images/videos every time. Not very indicative of training working so we want to ensure each generation starts with the same seed

finetrainers/hunyuan_video/hunyuan_video_lora.py
```diff
@@ -58,14 +58,15 @@ def load_latent_models(
...
+    shift: float = 1.0,
     revision: Optional[str] = None,
     cache_dir: Optional[str] = None,
     **kwargs,
 ) -> Dict[str, Union[nn.Module, FlowMatchEulerDiscreteScheduler]]:
     transformer = HunyuanVideoTransformer3DModel.from_pretrained(
         model_id, subfolder="transformer", torch_dtype=transformer_dtype, revision=revision, cache_dir=cache_dir
     )
-    scheduler = FlowMatchEulerDiscreteScheduler()
+    scheduler = FlowMatchEulerDiscreteScheduler(shift=shift)
```
a-r-r-o-w, line 69: HunyuanVideo uses 7.0 as the flow shift for inference. By default this value is 1.0, which corresponds to the original flow matching objective, but there has been reports of success and even better results with varying values of shift, so I think makes sense to support

finetrainers/dataset.py
```diff
@@ -20,13 +22,25 @@
...
+    PRECOMPUTED_DIR_NAME,
+    PRECOMPUTED_LATENTS_DIR_NAME,
+)
 
 
 logger = get_logger(__name__)
 
 
-class VideoDataset(Dataset):
+# TODO(aryan): This needs a refactor with separation of concerns.
```
a-r-r-o-w, line 36: Dataset part needs a complete rewrite. Will take it up in follow-up PRs

finetrainers/args.py
```diff
@@ -143,11 +144,24 @@ def to_dict(self) -> Dict[str, Any]:
...
                 "caption_dropout_technique": self.caption_dropout_technique,
                 "precompute_conditions": self.precompute_conditions,
+                "remove_common_llm_caption_prefixes": self.remove_common_llm_caption_prefixes,
             },
             "dataloader_arguments": {
                 "dataloader_num_workers": self.dataloader_num_workers,
                 "pin_memory": self.pin_memory,
             },
+            "diffusion_arguments": {
+                "flow_resolution_shifting": self.flow_resolution_shifting,
```
a-r-r-o-w, line 154: I still haven't had sucess with flow_resolution_shifting (needs to be added for LTX I believe), so we still don't handle the case of adjusting sigmas when this is specified. Will add the actual logic that makes use of it in the image-to-video PR after further iterations

a-r-r-o-w:
@sayakpaul Ready for another review. Doing a small run to check if the validation generator changes work as expected

finetrainers/trainer.py
```diff
@@ -971,15 +974,15 @@ def validate(self, step: int, final_validation: bool = False) -> None:
...
+            for index, (key, value) in enumerate(list(artifacts.items())):
                 artifact_type = value["type"]
                 artifact_value = value["value"]
                 if artifact_type not in ["image", "video"] or artifact_value is None:
                     continue
 
                 extension = "png" if artifact_type == "image" else "mp4"
                 filename = "validation-" if not final_validation else "final-"
-                filename += f"{step}-{accelerator.process_index}-{prompt_filename}.{extension}"
+                filename += f"{step}-{accelerator.process_index}-{index}-{prompt_filename}.{extension}"
```
a-r-r-o-w, line 985: Just caught my eye. If we use the same prompt multiple times, but want to validate at different resolutions, it might end up using the same filename I think. This should be safer imo

a-r-r-o-w:
I can confirm that the validation generator changes work as expected. Here's demo video using the same starting generator across different training steps:

https://github.com/user-attachments/assets/0aa2e6ab-3944-446f-bebf-cb452b71d73c

a-r-r-o-w:
@sayakpaul Proceeding with merge here as two folks who've helped with initial feedback wanted to try this with FP8 training. Currently, creating a common branch with these changes and the one for FP8 creates a merge conflict so I will resolve it for them in the FP8 branch. If you have any suggestions on things to improve here, happy to iterate in future PR

PR #176, opened 2025-01-03
support model cards

sayakpaul:
* Adds support for model cards.
* Allows to pass `fps`.
* Factors out artifacts generation and logging related utilities from `Trainer` to keep it leaner.


finetrainers/utils/hub_utils.py
```diff
@@ -0,0 +1,77 @@
...
+from typing import List, Union
+
+import numpy as np
+import wandb
+from diffusers.utils import export_to_video
+from diffusers.utils.hub_utils import load_or_create_model_card, populate_model_card
+from PIL import Image
+
+
+def save_model_card(
```
sayakpaul, line 11: Want to keep it minimal yet sufficient for now.

finetrainers/trainer.py
```diff
@@ -912,90 +909,33 @@ def validate(self, step: int, final_validation: bool = False) -> None:
...
-                    artifact_value.save(filename)
-                    artifact_value = wandb.Image(filename)
-                elif artifact_type == "video":
-                    logger.debug(f"Saving video to {filename}")
-                    # TODO: this should be configurable here as well as in validation runs where we call the pipeline that has `fps`.
-                    export_to_video(artifact_value, filename, fps=15)
-                    artifact_value = wandb.Video(filename, caption=prompt)
-
-                all_processes_artifacts.append(artifact_value)
+        all_processes_artifacts, prompts_to_filenames = generate_artifacts(
```
a-r-r-o-w, line 912: Please revert all changes here. It is not in the state to refactor and I would not do it just yet. Have other things in mind about improving this and how parameters are specified

finetrainers/trainer.py
```diff
@@ -912,90 +909,33 @@ def validate(self, step: int, final_validation: bool = False) -> None:
...
-                    image_artifacts = [artifact for artifact in all_artifacts if isinstance(artifact, wandb.Image)]
-                    video_artifacts = [artifact for artifact in all_artifacts if isinstance(artifact, wandb.Video)]
-                    tracker.log(
-                        {
-                            tracker_key: {"images": image_artifacts, "videos": video_artifacts},
-                        },
-                        step=step,
-                    )
-
+            log_artifacts(artifacts=all_artifacts, trackers=accelerator.trackers, tracker_key=tracker_key, step=step)
```
a-r-r-o-w, line 928: Only the model card related changes are fine to keep. No need to abstract until required (say, tensorboard support) or we have better tests. I don't think this PR was tested based on seeing the FPS changes

finetrainers/utils/artifact_utils.py
```diff
@@ -0,0 +1,156 @@
...
+    wandb_tracking = any("wandb" in tracker.name for tracker in trackers)
+    if wandb_tracking:
+        import wandb
+
+    all_processes_artifacts = []
+    num_validation_samples = len(args.validation_prompts)
+    prompts_to_filenames = {}
+    for i in range(num_validation_samples):
+        # Skip current validation on all processes but one
+        if i % num_processes != process_index:
```
a-r-r-o-w, line 39: Let me emphasize again that anything distributed should only remain in the trainer. Utility functions or helper functions should do only one thing.

I think let's limit this PR to model card changes and revert everything else.

finetrainers/args.py
```diff
@@ -626,6 +628,12 @@ def _add_validation_arguments(parser: argparse.ArgumentParser) -> None:
         default=False,
         help="Whether or not to enable model-wise CPU offloading when performing validation/testing to save memory.",
     )
+    parser.add_argument(
+        "--fps",
```
a-r-r-o-w, line 632: Have you tested this to work before asking for the review? Because currently it doesn't as result_args is never assigned in the function that handles pre-processing of the args. We only support training at one FPS (24) at the moment so allowing a custom value does not make sense either.

Only LTX uses it atm. Based on all design choices made so far, I fail to see why we add a statement to hunyuan and cog for passing fps when it won't be used.

Supporting multi-fps for training is not easily possible either because we don't preprocess any metadata. That is a fight for another day. Would revert any changes made for fps here.

a-r-r-o-w (commented):
Thanks!

finetrainers/utils/artifact_utils.py
sayakpaul: I will revert the changes but I didn't pass `accelerator` here. I thought it was okay to pass the arguments required to call a utility function in isolation. So, I don't know the deal here.

finetrainers/args.py
sayakpaul: > Have you tested this to work before asking for the review?

Sure, sorry for the bug. I acted in the moment, thinking `fps` would be a nice thing to let users configure without thinking about these. But we're only human beings, Aryan. So, perhaps we could tone down a bit to be a bit friendlier in our discussions. 

I could have been enraged after I found out the broken stuff which I had to fix in https://github.com/a-r-r-o-w/finetrainers/pull/177/. But I don't think that is a good way to work with a team spirit. We're not calling blue on blue here. 

Can we agree on that?

finetrainers/utils/artifact_utils.py
a-r-r-o-w: No. Making the utility handle distributed-ness by passing the accelerator indirectly (via the parameters here) is the same as just passing the accelerator. Utility functions should not deal with distributed-ness of any kind because it will make implementing a generic model parallelism harder in the future, and I want to limit such changes from coming into the codebase as well as remove any bad choices that already exist early on

sayakpaul: Makes sense, thanks for explaining. I will follow this through.

finetrainers/args.py
a-r-r-o-w: I mentioned here (https://github.com/a-r-r-o-w/finetrainers/pull/165#pullrequestreview-2528517092) that PR was okay to merge but just wanted to do a few long runs to make sure everything was working. There are many times, even in Diffusers, where we approve a PR because things look good and leave comments for what needs to be tested and wait until those have been addressed and we say "Yes, it's working as expected, so proceeding with merge" or something along those lines. I strongly believe we would have caught everything then if it was not merged before showing logs of a long run because:
- since noisy_latents were now in fp32, training time would have been a lot longer.
- i had just queued a LTX 1000-step run before going for a nap which errored-out immediately because of scheduler_alphas (only found out now that I woke up)

Did we do a long run before merging the PR to ensure that the training worked?

Apologies again for sounding rude. I need some work to be done on my online communication skills because even when I don't mean to be rude, it seems to be portrayed that way because I'm blunt and try to be to the point. Will make a conscious effort to do better

sayakpaul: > since noisy_latents were now in fp32, training time would have been a lot longer.

Well, it errored out because of type mismatches because `transformer` was in `torch.bfloat16`. There was no other issue.  

On the topic of sounding blunt and being to the point -- sure! At the risk of trying to bat up the past (my intention is to show that it doesn't hurt to make an effort here):

> https://github.com/a-r-r-o-w/finetrainers/pull/129#issuecomment-2559589003

I was trying to be to the point and direct here https://github.com/a-r-r-o-w/finetrainers/pull/129#issuecomment-2559556523 and realized it didn't reflect my regard of the points you conveyed. So, it was obvious to take it the other way.

sayakpaul:
Just tested the local changes on a small run: https://huggingface.co/sayakpaul/cog2b_disney (with https://github.com/a-r-r-o-w/finetrainers/pull/178 merged in)

LMK if there's anything else left off to be addressed in this PR.

finetrainers/args.py
a-r-r-o-w: The reason for my (rude) response there was not because you were blunt. It was because you took the simplest two examples out of what I tried conveying for what would logically define a trainer and what I envision it to be. 

In hindsight, maybe you're right and it's better to not have a separate trainer for full finetuning because of 95% similar logic. Will make the relevant changes to support LoRA and full through the same trainer. Hopefully, it's minimal changes.

On a separate note, am I right in understanding that there was no training run done before the Cog PR was merged (after the latest changes)? Which would mean, we don't know if the training actually works yet. I'm queuing up some runs now based on the things I wanted to test in my comment so will be on the lookout for anything else that is broken

sayakpaul: Sounds good to me!

a-r-r-o-w: That is an ambiguous response to my question and I'm not sure how to interpret it. If yes, can I please have access to the logs so am convinced that it worked before the PR was merged? If not, let's make sure to have atleast one significant > 1000 step run for any models added going forward

sayakpaul: Sorry. I am okay with you trying 

> I'm queuing up some runs now based on the things I wanted to test in my comment so will be on the lookout for anything else that is broken

We can split the experiments if needed as well. LMK. I will leave it to you to merge the PR if rest of the changes look good to you AND we're convinced the runs are effective. 

Does that sound good? A significant run wasn't done before the [Cog PR](https://github.com/a-r-r-o-w/finetrainers/pull/165) was merged.

a-r-r-o-w (approved):
Thank you, the changes look good. I haven't run any tests to verify but the I'm assuming [this](https://huggingface.co/sayakpaul/cog2b_disney) comes from changes introduced here so looks good to merge. Okay to merge if you feel this is good to go, thanks!

sayakpaul:
> I haven't run any tests to verify but the I'm assuming [this](https://huggingface.co/sayakpaul/cog2b_disney) comes from changes introduced here so looks good to merge.

Yes, that is correct.

Would you rather have me run some complete runs for Cog (2b, 5b)? Completely fine by me if so.

finetrainers/trainer.py
```diff
@@ -988,6 +991,8 @@ def validate(self, step: int, final_validation: bool = False) -> None:
                     # TODO: this should be configurable here as well as in validation runs where we call the pipeline that has `fps`.
                     export_to_video(artifact_value, filename, fps=15)
                     artifact_value = wandb.Video(filename, caption=prompt)
+                    if accelerator.is_main_process:
+                        prompts_to_filenames[prompt] = filename
```
a-r-r-o-w, line 995: So just tested it and it seems like here the filename ends up becoming something like `/raid/aryan/final-validation-...`. This is because we join the output_dir to the filename when saving the video

image

So, I think we need to make sure we use the filename before it gets prefixed with output_dir

a-r-r-o-w:
The training seems to work phenomally, so all changes related to Cog seem good! https://wandb.ai/aryanvs/finetrainers-cogvideox/runs/6925e91q. It seems like I used too high a learning rate because the validation videos collapses a bit in some runs, but definitely seems like the model is learning and there is no quality degradation.

> Would you rather have me run some complete runs for Cog (2b, 5b)? Completely fine by me if so.

Completely up to you if you'd like to release something for a specific style! I did a 1000 step run for Cakify so maybe you could check with Disney dataset if you'd like. My experiments are more aimed towards figuring out if SNR loss weighting has been implemented correctly in the original Cog codebase as well as our scripts. Will probably be easier to figure out with full finetuning so will put on hold and prioritize memory optimizations

finetrainers/trainer.py
sayakpaul: Hmm not sure about it.

I inspected the `output_dir`:

```bash
ls /raid/.cache/huggingface/sayak/cog2b/cog2b_disney
checkpoint-10                             pytorch_lora_weights.safetensors               validation-5-0-BW_STYLE-A-black-and-whit.mp4
checkpoint-5                              README.md                                      validation-5-1-A-woman-with-long-brown-h.mp4
final-10-0-BW_STYLE-A-black-and-whit.mp4  validation-10-0-BW_STYLE-A-black-and-whit.mp4
final-10-1-A-woman-with-long-brown-h.mp4  validation-10-1-A-woman-with-long-brown-h.mp4
```

Seems alright no?

Also https://huggingface.co/sayakpaul/cog2b_disney seems alright in the sense that the resultant model card looks as expected.

sayakpaul:
@a-r-r-o-w thanks for the positive note and testing that!

I think I will merge this PR given you have verified it. But before that I will wait for your response to https://github.com/a-r-r-o-w/finetrainers/pull/176#discussion_r1903261392.

finetrainers/trainer.py
a-r-r-o-w: I mean to say that the model card does not render the demo video because of the incorrect path to the video relative to the uploaded repository (the case with your repo as well).

In the created model card, the path shouldn't contain the full path with output_dir (for example, right now it creates something like /raid/aryan/cogvideox-lora/validation-video.mp4). But for the model card, it should just be validation-video.mp4)

sayakpaul: Just fixed.

https://huggingface.co/sayakpaul/cog2b_disney

sayakpaul:
Sorry for requesting the review again but just want to be sure. Have tested in https://github.com/a-r-r-o-w/finetrainers/pull/176/commits/800a47fb17190a0cc71bb18470ba345ed7953be3.

a-r-r-o-w (approved):

PR #180, opened 2025-01-04
[CI] add a workflow to do quality checks.

sayakpaul:
Follow-up would be a basic test suite (I hope you're okay with that, but if not, I will understand).

### DIFF
--- a/.github/workflows/pr_tests.yml
+++ b/.github/workflows/pr_tests.yml
@@ -1,0 +1,30 @@
+name: Fast tests for PRs
+
+on:
+  pull_request:
+    branches:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  check_code_quality:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.8"
+      - name: Install dependencies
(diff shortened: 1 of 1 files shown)


a-r-r-o-w (approved):
Thanks, this is great!

> Follow-up would be a basic test suite 

Yes, this is perfectly okay with me. To start, let's use dummy checkpoints from `hf-internal-testing` if possible because using the large original checkpoints will take up to 30 minutes to test everything.

Akin to slow tests in diffusers, we can test bigger checkpoints too, but only when introducing major changes. Small changes and bug fixes just need to ensure the quick small model tests pass. If you have other ideas on how to go about this, would love to hear. No strong opinions here from me

sayakpaul:
We're perfectly aligned on this!

> Yes, this is perfectly okay with me. To start, let's use dummy checkpoints from hf-internal-testing if possible because using the large original checkpoints will take up to 30 minutes to test everything.

Yes 100 percent. This will be our fast tests (i.e., running with dummy checkpoints and as such they will be smoke tests).

> Akin to slow tests in diffusers, we can test bigger checkpoints too, but only when introducing major changes. Small changes and bug fixes just need to ensure the quick small model tests pass. If you have other ideas on how to go about this, would love to hear. No strong opinions here from me

Yes! So, for now, I am thinking of just having a combination of a shell script and `Makefile` that we (you and I) will run manually on GPUs as needed based the criticality of the changes. This will allow us to fire all the commands / runs we need for a relatively quick validation (like the ones you suggested before).

Expect the PRs soon. 

Will merge the PR now that I have an approval.

PR #181, opened 2025-01-04
Added parameter description

ootsuka-repos:


sayakpaul:
Thanks for this PR!

Since we provide `help` for almost all the important arguments, I am wondering if using `train.py -h` would help the users know all about them? 

Perhaps we could just add a note about that? @a-r-r-o-w WDYT?

PR #189, opened 2025-01-07
Update requirements

### LINKED ISSUE #186: requirements torch<2.5.0 vs torch 2.5.1
### System Info / 系統信息

you mention in the readme that torch 2.5.1 or higher is recommended but in the requirements.txt you ask for torch<2.5.0.

### Information / 问题信息

- [ ] The official example scripts / 官方的示例脚本
- [ ] My own modified scripts / 我自己修改的脚本和任务

### Reproduction / 复现过程

.

### Expected behavior / 期待表现

.

a-r-r-o-w:
Fix #186


requirements.txt
```diff
@@ -1,16 +1,16 @@
 accelerate
 bitsandbytes
-diffusers>=0.30.3
+diffusers>=0.32.1
```
sayakpaul, line 3: Should we always tell users to install from the `main`. I think we do tell that in the README.

requirements.txt
```diff
@@ -1,16 +1,16 @@
...
+peft>=0.13.0
 decord>=0.6.0
 wandb
 pandas
-torch<2.5.0
-torchvision<0.20.0
-torchao>=0.5.0
+torch>=2.5.1
+torchvision>=0.20.1
+torchao>=0.7.0
```
sayakpaul, line 13: Not for this PR. But I would like to make it so that `torchao` is installed only when it's required by the user. I will work on it.

sayakpaul (approved):
Just a tiny comment but otherwise, LGTM!

requirements.txt
a-r-r-o-w: Sounds good

requirements.txt
a-r-r-o-w: Yep, that works. We currently don't support torchao in the new API either, so can be addressed together in near future

Issue #194, opened 2025-01-07
Call to nonexistent torch.mps.is_available() function


balbit:
### Minor Bug:

[This line of code](https://github.com/a-r-r-o-w/finetrainers/blob/main/finetrainers/utils/memory_utils.py#L24) checking for torch.mps availability calls torch.mps.is_available(), which is not part of the torch.mps [API](https://pytorch.org/docs/stable/mps.html)

This causes confusing errors when a cuda device is not provided.

### Information / 问题信息

- [X] The official example scripts / 官方的示例脚本
- [ ] My own modified scripts / 我自己修改的脚本和任务

### Reproduction / 复现过程

Run the official example script with LTX, without a cuda device available

### Expected behavior / 期待表现

Fail when cuda is unavailable, or detect mps memory.

a-r-r-o-w:
Oh, yes you're right. This should have been `torch.backends.mps.is_available()`. Would you be able to open a PR for fixing? Is there any other issues you face when training on Mac?

ArEnSc:
@balbit 
https://github.com/a-r-r-o-w/finetrainers/pull/195
Test this PR

balbit:
Yeah this works!

I'm not training on a Mac right now, just noticed this issue when running when cuda wasn't available and wanted it to display the right error message

PR #195, opened 2025-01-07
Fix minor bug with function call that doesn't exist.

ArEnSc:
https://github.com/a-r-r-o-w/finetrainers/issues/194


a-r-r-o-w (approved):
Thanks!

PR #198, opened 2025-01-09
Better defaults for LTXV

a-r-r-o-w:
The major change is that we using logit_normal weighting as described in the paper. We did not know this detail beforehand when LTXV support was added leading to suboptimal loss.


sayakpaul (approved):
Thanks!

We can continue to revisit these defaults (particularly these weighting schemes) as details become available to us. Currently, we have it for LTX (already done in this PR), Cog (done), and Mochi-1 (through their finetuner). 

Perhaps, it might be nice to have a check with the Hunyuan folks about this.

PR #192, opened 2025-01-07
Full Finetuning for LTX possibily extended to other models.

ArEnSc:
This pull request introduces the following changes:
Added a --supervised-finetuning  flag "sft" for each model to control fine-tuning individually.
Refactored To-Do items related to:
Managing the list of models.
Enabling gradients in the model components.

Todo: 
Run SFT for a few runs on one model to ensure stability

PR #202, opened 2025-01-09
Support full fine-tuning auxiliaries part ii

sayakpaul:
Tested with all three currently supported models.

### DIFF
--- a/finetrainers/trainer.py
+++ b/finetrainers/trainer.py
@@ -21,6 +21,7 @@ from accelerate.utils import (
     gather_object,
     set_seed,
 )
+from diffusers import DiffusionPipeline
 from diffusers.configuration_utils import FrozenDict
 from diffusers.models.autoencoders.vae import DiagonalGaussianDistribution
 from diffusers.optimization import get_scheduler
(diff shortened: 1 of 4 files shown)


finetrainers/trainer.py
```diff
@@ -910,35 +925,7 @@ def validate(self, step: int, final_validation: bool = False) -> None:
         memory_statistics = get_memory_statistics()
         logger.info(f"Memory before validation start: {json.dumps(memory_statistics, indent=4)}")
 
-        if not final_validation:
```
sayakpaul, line 913: From [here](https://github.com/a-r-r-o-w/finetrainers/pull/192#issuecomment-2579187405):

> Validation should not load lora weights, so needs to be done differently

The changes we see here are my understanding of doing the above. If it's not, I won't mind direct pushes to this branch to iterate faster.

a-r-r-o-w:
Could this be opened against #192 branch so that all changes could be merged together in one go?

finetrainers/trainer.py
```diff
@@ -459,33 +468,39 @@ def load_model_hook(models, input_dir):
...
-            incompatible_keys = set_peft_model_state_dict(transformer_, transformer_state_dict, adapter_name="default")
-            if incompatible_keys is not None:
-                # check only for unexpected keys
-                unexpected_keys = getattr(incompatible_keys, "unexpected_keys", None)
-                if unexpected_keys:
-                    logger.warning(
-                        f"Loading adapter weights from state_dict led to unexpected keys not found in the model: "
-                        f" {unexpected_keys}. "
+                if self.args.training_type == "lora":
+                    transformer_ = transformer_cls_.__class__.from_pretrained(
```
a-r-r-o-w, line 474: Maybe missing some context here, but do we need `__class__` on the `transformer_cls_` class ref? No problem even if not required since you've tested this to work already. Asking because, in the other branch, just `transformer_cls_.from_pretrained` is used

finetrainers/trainer.py
a-r-r-o-w: Changes look good!

a-r-r-o-w (approved):
LGTM. But maybe let's open PR in common branch and merge with @ArEnSc's PR

finetrainers/trainer.py
sayakpaul: You're right.

sayakpaul:
Closing as I pushed my updates to https://github.com/a-r-r-o-w/finetrainers/pull/192

PR #204, opened 2025-01-09
Shell script to minimally test supported models on a real dataset

sayakpaul:
Have run and can confirm it's working as expected. Since this script will be used quite (or so I think) heavily, I would appreciate if you could run it, too.

### DIFF
--- a/tests/scripts/dummy_cogvideox_lora.sh
+++ b/tests/scripts/dummy_cogvideox_lora.sh
@@ -78,4 +78,4 @@ eval $cmd
 echo -ne "-------------------- Finished executing script --------------------\n\n"
 
 rm -rf $OUTPUT_DIR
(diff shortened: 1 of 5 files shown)


a-r-r-o-w (approved):
Thanks Sayak! The changes look good.

I'm always a bit skeptic about using `rm -rf` in scripts, but it's alright for now. I will take up some future test additions to run these from pytest (similar to our test lora scripts), so that we can use TemporaryDirectory() and all the python goodies

sayakpaul:
> I will take up some future test additions to run these from pytest (similar to our test lora scripts), so that we can use TemporaryDirectory() and all the python goodies

Yes, I am working on testing with dummy small checkpoints in a formal manner i.e., like we do in Diffusers.

PR #215, opened 2025-01-13
Gradient update fixes

a-r-r-o-w:


a-r-r-o-w:
Tested, and seems like not needed because accelerate handles this internally:
- https://github.com/huggingface/accelerate/blob/f0b030554cbcd01c5541c449e92066715f21a99e/src/accelerate/optimizer.py#L161

PR #223, opened 2025-01-15
[WIP][feat] add Mochi T2V

sayakpaul:
Fixes #90 #217 

This PR is not at all in a complete state. I am opening this PR in a minimal capacity to discuss some implementation details. 

Once those are sorted, I will ground it to complementation and for a more rigorous review.

My questions are commented in the changes I am introducing.

### DIFF
--- a/finetrainers/models/mochi/lora.py
+++ b/finetrainers/models/mochi/lora.py
@@ -21,7 +21,7 @@ def cast_dit(model, dtype):
 
 def load_condition_models(
     model_id: str = "genmo/mochi-1-preview",
-    text_encoder_dtype: torch.dtype = torch.float32,
+    text_encoder_dtype: torch.dtype = torch.bfloat16,
     revision: Optional[str] = None,
     cache_dir: Optional[str] = None,
     **kwargs,
@@ -35,7 +35,7 @@ def load_condition_models(
 
 def load_latent_models(
     model_id: str = "genmo/mochi-1-preview",
-    vae_dtype: torch.dtype = torch.float32,
+    vae_dtype: torch.dtype = torch.bfloat16,
     revision: Optional[str] = None,
     cache_dir: Optional[str] = None,
     **kwargs,
--- a/finetrainers/models/mochi/lora.py
+++ b/finetrainers/models/mochi/lora.py
@@ -1,0 +1,281 @@
+from typing import Any, Dict, List, Optional, Union
+
+import torch
+from diffusers import AutoencoderKLMochi, FlowMatchEulerDiscreteScheduler, MochiPipeline, MochiTransformer3DModel
+from PIL import Image
+from transformers import T5EncoderModel, T5Tokenizer
+
+
+# Following the official implementation.
+def cast_dit(model, dtype):
+    for name, module in model.named_modules():
+        if isinstance(module, torch.nn.Linear):
+            assert any(
+                n in name for n in ["time_embed", "proj_out", "blocks", "norm_out"]
+            ), f"Unexpected linear layer: {name}"
+            module.to(dtype=dtype)
+        elif isinstance(module, torch.nn.Conv2d):
+            module.to(dtype=dtype)
+    return model
+
+
+def load_condition_models(
+    model_id: str = "genmo/mochi-1-preview",
+    text_encoder_dtype: torch.dtype = torch.float32,
+    revision: Optional[str] = None,
+    cache_dir: Optional[str] = None,
+    **kwargs,
+):
+    tokenizer = T5Tokenizer.from_pretrained(model_id, subfolder="tokenizer", revision=revision, cache_dir=cache_dir)
+    text_encoder = T5EncoderModel.from_pretrained(
+        model_id, subfolder="text_encoder", torch_dtype=text_encoder_dtype, revision=revision, cache_dir=cache_dir
+    )
+    return {"tokenizer": tokenizer, "text_encoder": text_encoder}
+
+
+def load_latent_models(
+    model_id: str = "genmo/mochi-1-preview",
+    vae_dtype: torch.dtype = torch.float32,
+    revision: Optional[str] = None,
+    cache_dir: Optional[str] = None,
+    **kwargs,
+):
+    vae = AutoencoderKLMochi.from_pretrained(
+        model_id, subfolder="vae", torch_dtype=vae_dtype, revision=revision, cache_dir=cache_dir
+    )
+    return {"vae": vae}
+
+
+def load_diffusion_models(
+    model_id: str = "genmo/mochi-1-preview",
+    transformer_dtype: torch.dtype = torch.float32,
+    revision: Optional[str] = None,
+    cache_dir: Optional[str] = None,
+    **kwargs,
+):
+    transformer = MochiTransformer3DModel.from_pretrained(
+        model_id, subfolder="transformer", torch_dtype=transformer_dtype, revision=revision, cache_dir=cache_dir
+    )
+    transformer = cast_dit(transformer, torch.bfloat16)
+    scheduler = FlowMatchEulerDiscreteScheduler.from_pretrained(model_id, subfolder="scheduler")
+    return {"transformer": transformer, "scheduler": scheduler}
+
+
+def initialize_pipeline(
+    model_id: str = "genmo/mochi-1-preview",
+    text_encoder_dtype: torch.dtype = torch.float32,
+    transformer_dtype: torch.dtype = torch.float32,
+    vae_dtype: torch.dtype = torch.float32,
+    tokenizer: Optional[T5Tokenizer] = None,
+    text_encoder: Optional[T5EncoderModel] = None,
+    transformer: Optional[MochiTransformer3DModel] = None,
+    vae: Optional[AutoencoderKLMochi] = None,
+    scheduler: Optional[FlowMatchEulerDiscreteScheduler] = None,
+    device: Optional[torch.device] = None,
+    revision: Optional[str] = None,
+    cache_dir: Optional[str] = None,
+    enable_slicing: bool = False,
+    enable_tiling: bool = False,
+    enable_model_cpu_offload: bool = False,
+    is_training: bool = False,
+    **kwargs,
+) -> MochiPipeline:
+    component_name_pairs = [
+        ("tokenizer", tokenizer),
+        ("text_encoder", text_encoder),
+        ("transformer", transformer),
+        ("vae", vae),
+        ("scheduler", scheduler),
+    ]
+    components = {}
+    for name, component in component_name_pairs:
+        if component is not None:
+            components[name] = component
+
+    pipe = MochiPipeline.from_pretrained(model_id, **components, revision=revision, cache_dir=cache_dir)
+    pipe.text_encoder = pipe.text_encoder.to(dtype=text_encoder_dtype)
+    pipe.vae = pipe.vae.to(dtype=vae_dtype)
+
+    # The transformer should already be in the correct dtype when training, so we don't need to cast it here.
+    # If we cast, whilst using fp8 layerwise upcasting hooks, it will lead to an error in the training during
+    # DDP optimizer step.
+    if not is_training:
+        pipe.transformer = pipe.transformer.to(dtype=transformer_dtype)
+
+    if enable_slicing:
+        pipe.vae.enable_slicing()
+    if enable_tiling:
+        pipe.vae.enable_tiling()
+
+    if enable_model_cpu_offload:
+        pipe.enable_model_cpu_offload(device=device)
+    else:
+        pipe = pipe.to(device=device)
+
+    return pipe
+
+
+def prepare_conditions(
+    tokenizer,
+    text_encoder,
+    prompt: Union[str, List[str]],
+    device: Optional[torch.device] = None,
+    dtype: Optional[torch.dtype] = None,
+    max_sequence_length: int = 256,  # TODO: this should be configurable
+    **kwargs,
+):
+    device = device or text_encoder.device
+    dtype = dtype or text_encoder.dtype
+    return _get_t5_prompt_embeds(
+        tokenizer=tokenizer,
+        text_encoder=text_encoder,
+        prompt=prompt,
+        max_sequence_length=max_sequence_length,
+        device=device,
+        dtype=dtype,
+    )
+
+
+# Following original implementation.
+@torch.autocast("cuda", dtype=torch.bfloat16)
+@torch.inference_mode()
+def prepare_latents(
+    vae: AutoencoderKLMochi,
+    image_or_video: torch.Tensor,
+    device: Optional[torch.device] = None,
+    dtype: Optional[torch.dtype] = None,
+    generator: Optional[torch.Generator] = None,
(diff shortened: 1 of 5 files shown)


finetrainers/models/mochi/lora.py
```diff
@@ -0,0 +1,281 @@
...
+    model_id: str = "genmo/mochi-1-preview",
+    transformer_dtype: torch.dtype = torch.float32,
+    revision: Optional[str] = None,
+    cache_dir: Optional[str] = None,
+    **kwargs,
+):
+    transformer = MochiTransformer3DModel.from_pretrained(
+        model_id, subfolder="transformer", torch_dtype=transformer_dtype, revision=revision, cache_dir=cache_dir
+    )
+    transformer = cast_dit(transformer, torch.bfloat16)
```
sayakpaul, line 59: If we do decide to go forward with this (the original implementation does it), this should be guarded with `is_training`, I think. Because when not training, we don't need to do this.

finetrainers/models/mochi/lora.py
```diff
@@ -0,0 +1,281 @@
...
+    scheduler = FlowMatchEulerDiscreteScheduler.from_pretrained(model_id, subfolder="scheduler")
+    return {"transformer": transformer, "scheduler": scheduler}
+
+
+def initialize_pipeline(
+    model_id: str = "genmo/mochi-1-preview",
+    text_encoder_dtype: torch.dtype = torch.float32,
+    transformer_dtype: torch.dtype = torch.float32,
+    vae_dtype: torch.dtype = torch.float32,
```
sayakpaul, line 68: Everything in FP32, following the original implementation and relying on autocast instead.

finetrainers/models/mochi/lora.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+def collate_fn_t2v(batch: List[List[Dict[str, torch.Tensor]]]) -> Dict[str, torch.Tensor]:
+    return {
+        "prompts": [x["prompt"] for x in batch[0]],
+        "videos": torch.stack([x["video"] for x in batch[0]]),
+    }
+
+
+# Following original implementation.
```
sayakpaul, line 189: Original implementation plus I am not sure how we are dealing with making `accelerate` aware that we want to do `mixed_precision` training as we no longer set that attribute in `Accelerator()`.

finetrainers/models/mochi/lora.py
```diff
@@ -0,0 +1,281 @@
...
+        encoder_hidden_states=prompt_embeds,
+        encoder_attention_mask=prompt_attention_mask,
+        timestep=timesteps,
+        return_dict=False,
+    )[0]
+
+    return {"latents": denoised_latents}
+
+
+def validation(
```
sayakpaul, line 212: This can be revisited to match what we do in:
https://huggingface.co/docs/diffusers/main/en/api/pipelines/mochi#reproducing-the-results-from-the-genmo-mochi-repo

finetrainers/models/mochi/lora.py
```diff
@@ -0,0 +1,281 @@
...
+    transformer_dtype: torch.dtype = torch.float32,
+    revision: Optional[str] = None,
+    cache_dir: Optional[str] = None,
+    **kwargs,
+):
+    transformer = MochiTransformer3DModel.from_pretrained(
+        model_id, subfolder="transformer", torch_dtype=transformer_dtype, revision=revision, cache_dir=cache_dir
+    )
+    transformer = cast_dit(transformer, torch.bfloat16)
+    scheduler = FlowMatchEulerDiscreteScheduler.from_pretrained(model_id, subfolder="scheduler")
```
a-r-r-o-w, line 60: Would instead initialize the scheduler simply as `FlowMatchEulerDiscreteScheduler()`. This is so that the `invert_sigmas` flag in the original config does not mess with any of the downstream reversing of sigmas we apply.

We want the scheduler to be the original flow match objective, and make modifications specific to Mochi in the utility functions

finetrainers/models/mochi/lora.py
a-r-r-o-w: Since `load_diffusion_models` is only called for training purposes, there's no need to guard here I think. Only `initialize_pipeline` needs to be guarded, which is specific to layerwise fp8 upcasting case

a-r-r-o-w (commented):
Thanks for looking into this, Sayak.

I've made some suggestions about how we can handle the reversed timesteps. I'm not sure if they'd work 100% out of the box, so we can only know by trying some runs. Maybe we don't need any changes to the flow matching algorithm itself, but just what the prediction returns and the reversed timestep to condition on.

Most of the changes look good. I have a few concerns though:
- Personally never prefer using torch.autocast anywhere. Okay to keep if you strongly believe it's required to make lora training work as expected
- Would default the text encoder and vae to `torch.bfloat16` if we can test that training works with that. I'm not sure text encoder and vae precision has significant impact on Mochi quality, but I'd like to believe it does not (atleast for bf16). This is because the default options should lead to "okay" memory usage. Since we allow configuration of these dtypes via the command line, they can be made fp32 if user wants
- `cast_dit` is okay to keep, but would maybe train two loras to test if it is required by default (which may marginally increase memory requirements) - one with transformer in pure bf16 and one with cast_dit that keeps some layers in fp32

sayakpaul:
This is perfect feedback. Will get to them tomorrow, thanks Aryan!

finetrainers/utils/diffusion_utils.py
```diff
@@ -134,9 +137,12 @@ def prepare_target(
     scheduler: Union[CogVideoXDDIMScheduler, FlowMatchEulerDiscreteScheduler],
     noise: torch.Tensor,
     latents: torch.Tensor,
+    is_mochi: bool = False,
 ) -> torch.Tensor:
     if isinstance(scheduler, FlowMatchEulerDiscreteScheduler):
         target = noise - latents
+        if is_mochi:
+            target = -target
```
sayakpaul, line 145:
```suggestion
```

finetrainers/utils/diffusion_utils.py
```diff
@@ -93,6 +93,9 @@ def prepare_sigmas(
     device: torch.device = torch.device("cpu"),
     generator: Optional[torch.Generator] = None,
 ) -> torch.Tensor:
+    # TODO: Mochi only does: sigmas = torch.rand(latents.shape[0])
+    # It doesn't rely on `sigmas` configured in the scheduler. To handle that, should
+    # Mochi implement its own `prepare_sigmas()` similar to how `calculate_noisy_latents()` is implemented?
```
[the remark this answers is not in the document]
sayakpaul: So, I emulated something similar:

```py
from diffusers import FlowMatchEulerDiscreteScheduler
from finetrainers.utils.diffusion_utils import get_scheduler_sigmas, prepare_sigmas
import torch 

batch_size = 2

scheduler = FlowMatchEulerDiscreteScheduler()
scheduler_sigmas = get_scheduler_sigmas(scheduler)
sigmas = prepare_sigmas(
    scheduler=scheduler,
    sigmas=scheduler_sigmas,
    batch_size=batch_size,
    num_train_timesteps=scheduler.config.num_train_timesteps,
    flow_weighting_scheme=None,
    device="cpu",
    generator=torch.manual_seed(0),
)
print(sigmas)

print(torch.rand(2, generator=torch.manual_seed(0)))
```

Prints:

```bash
tensor([0.5040, 0.2320])
tensor([0.4963, 0.7682])
```

In case of "none" weighting scheme, we apply uniform weighting through `torch.randn()` and then compute the sigmas with indexing. But this is a bit different I guess.

finetrainers/models/mochi/lora.py
```diff
@@ -0,0 +1,294 @@
...
+        encoder_attention_mask=prompt_attention_mask,
+        # TODO: revisit if needed as Mochi has a weird way of doing `timesteps`.
+        timestep=scheduler.config.num_train_timesteps - timesteps,
+        return_dict=False,
+    )[0]
+    # TODO: revisit if needed. We do this because of 
+    # https://github.com/genmoai/mochi/blob/aba74c1b5e0755b1fa3343d9e4bd22e89de77ab1/src/genmo/mochi_preview/dit/joint_model/asymm_models_joint.py#L656
+    # In short, Mochi operates on reversed targets which is why we need to negate 
+    # the predictions.
+    denoised_latents = -denoised_latents
```
sayakpaul, line 220: For our reference:

```py
import torch 
import torch.nn.functional as F

latents = torch.randn(1, 4, 64, 64, generator=torch.manual_seed(0))
noise = torch.randn(1, 4, 64, 64, generator=torch.manual_seed(1))
denoised_latents = torch.randn(1, 4, 64, 64, generator=torch.manual_seed(2))

# mochi way
targets = latents - noise 
print("Mochi Way Loss:", F.mse_loss(denoised_latents.float(), targets.float()))

# our way
target = noise - latents  # Correctly set the target for "our way"
denoised_latents = -denoised_latents  # Negate the denoised_latents
print("Our Way Loss:", F.mse_loss(denoised_latents.float(), target.float()))  # Compare to the correct target

```

Yields same results.

finetrainers/utils/diffusion_utils.py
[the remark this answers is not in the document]
a-r-r-o-w: I don't think we apply uniform weighting through `torch.randn`, no? We apply it via `torch.rand` too [here](https://github.com/a-r-r-o-w/finetrainers/blob/f5ecb6e4711ea0f0b27992e36cc7a72dfc12fae1/finetrainers/utils/diffusion_utils.py#L62).

From the uniformly sampled random values, we create indices on the linspace sigmas, so I think it should be the same effect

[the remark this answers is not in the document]
sayakpaul: My bad. I meant `torch.rand()`, only. 

The effect should be the same but I want to be sure so I will first try out "weighting_scheme=None" and then try out what's originally done in Mochi.

sayakpaul:
Some more investigations:

```py
from diffusers import MochiTransformer3DModel, FlowMatchEulerDiscreteScheduler
from finetrainers.utils.diffusion_utils import prepare_sigmas
import torch


transformer = MochiTransformer3DModel.from_pretrained(
    "genmo/mochi-1-preview", subfolder="transformer"
).to("cuda")
scheduler = FlowMatchEulerDiscreteScheduler()

latents = torch.randn(1, 12, 4, 60, 106, generator=torch.manual_seed(0)).to("cuda")
prompt_embeds = torch.randn(1, 256, 4096, generator=torch.manual_seed(0)).to("cuda")
prompt_attention_mask = torch.randint(0, 2, size=(1, 256), generator=torch.manual_seed(0)).to("cuda")

# Original way.
# We are initializing sigma on CPU and then moving to CUDA.
original_sigma = torch.rand(latents.shape[0], generator=torch.manual_seed(0)).to("cuda")
timesteps = ((1 - original_sigma) * scheduler.config.num_train_timesteps).long()

with torch.no_grad():
    model_pred = transformer(
        hidden_states=latents,
        encoder_hidden_states=prompt_embeds,
        encoder_attention_mask=prompt_attention_mask,
        timestep=timesteps,
        return_dict=False,
    )[0]
    print(model_pred.shape)


# Our way.
scheduler_sigmas = scheduler.sigmas.clone()
sigmas = prepare_sigmas(
    scheduler=scheduler,
    sigmas=scheduler_sigmas,
    batch_size=1,
    num_train_timesteps=scheduler.config.num_train_timesteps,
    flow_weighting_scheme="none",
    device="cpu",
    generator=torch.manual_seed(0),
)
timesteps = (sigmas * scheduler.config.num_train_timesteps).long()
timesteps = timesteps.to("cuda")

with torch.no_grad():
    our_model_pred = transformer(
        hidden_states=latents,
        encoder_hidden_states=prompt_embeds,
        encoder_attention_mask=prompt_attention_mask,
        timestep=scheduler.config.num_train_timesteps - timesteps,
        return_dict=False,
    )[0]
    print(our_model_pred.shape)

try:
    print(torch.testing.assert_close(model_pred, our_model_pred, atol=1e-4, rtol=1e-4))
except AssertionError:
    print("Nope :()")

# If we combine original way of computing sigma and our way of getting timesteps?
timesteps = (original_sigma * scheduler.config.num_train_timesteps).long()
with torch.no_grad():
    hybrid_model_pred = transformer(
        hidden_states=latents,
        encoder_hidden_states=prompt_embeds,
        encoder_attention_mask=prompt_attention_mask,
        timestep=scheduler.config.num_train_timesteps - timesteps,
        return_dict=False,
    )[0]
    print(our_model_pred.shape)

try:
    print(torch.testing.assert_close(model_pred, hybrid_model_pred, atol=1e-4, rtol=1e-4))
except AssertionError:
    print("model_pred, hybrid_model_pred didn't.")

try:
    print(torch.testing.assert_close(our_model_pred, hybrid_model_pred, atol=1e-4, rtol=1e-4))
except AssertionError:
    print("our_model_pred, hybrid_model_pred didn't.")

```

None of the assertions pass here.

I investigated the `timesteps` calculations a bit more:

```py
import torch

# original way
sigma = torch.rand(1, generator=torch.manual_seed(0))
timestep = ((1 - sigma) * 1000).long()

# our way
a_timestep = ((sigma) * 1000).long()
a_timestep = 1000 - a_timestep

print(timestep) # prints 503
print(a_timestep) # prints 504
```

I know the effects are some but there is a difference that is worth noting here.

finetrainers/models/mochi/lora.py
a-r-r-o-w: Perfect 👨‍🍳

sayakpaul:
Yes, I am not worried about it but noting here in case things go south hahaha.

a-r-r-o-w:
@sayakpaul Will this be going into the release by any chance? Or would you like to continue work on it after the parallel PR is merged (which will be breaking in regard to current design)?

sayakpaul:
Will follow your parallel PR to work on it. This way I will make myself familiarize with your code and get it done. WDYT?

a-r-r-o-w:
Perfect, sounds good! Feel free to leave any comments on the PR and I'll try to answer them. We currently support the following:
- DDP (`dp_degree`)
- FSDP (`dp_shards`)
- HSDP (both `dp_degree` and `dp_shards` > 1)
- TP (`tp_degree`, but I still need to write/figure-out tensor parallel plans for each model with minimal amount of comms)

Handling model saving is remaining, which I'll add as soon as I'm finished testing pipeline parallel (`pp_degree`). We have two backends for now:
- Accelerate: Only supports DDP for now. I'll write the FSDP parts soon
- Pytorch DTensor: Adapted from torchtune and torchtitan (basically torch API). Supports all of the above

The other backends are quite incomplete at the moment since I haven't had the time to complete, but have made significant progress. The only stable backends we'll try and support are Accelerate & PTD (the others are just for satisfying my curiosity and benchmarking against). I'll make sure to write clear and concise docs explaining all of this.

My priority for next backend is ParaAttention followed by DeepSpeed native, since I'm in touch with chengzeyi and he's super happy to help us with any issues! These are all handled in their standalone files that users will not have to worry about.

We also do precomputation in a new way so as to not hog up disk space for large datasets. I believe this to be one of the most impactful features in the PR. Will write explanations for everything in that PR very soon!

Issue #222, opened 2025-01-15
How does `accelerate` use `mixed_precision`


sayakpaul:
@a-r-r-o-w I see that we no longer set the `mixed_precision` argument while initializing `Accelerator`:

https://github.com/a-r-r-o-w/finetrainers/blob/4271b4e436b6057c34c8cc155b997bb63f65fb00/finetrainers/trainer.py#L1001-L1006

How do we make use of mixed-precision training then?

a-r-r-o-w:
The config files still contain `mixed_precision`, and accelerate makes use of that if specified. We don't need to explicitly pass that from our end because I find it rather confusing to have `transformer_dtype` and `mixed_precision` together after the latest changes.

`transformer_dtype` should be the ultimate source of truth for the precision to be used when training. It will also most likely always have to be `bf16` because:
- All models currently supported (except Cog-2b maybe) do not work well in FP16 for inference, so training would be broken as well. We can revisit if it makes sense to train in fp16 for other models added, but again, we don't really have to pass anything from our end since accelerate will read from config file anyway
- No one really will be using fp32 for memory efficient training

So, the accelerate config files now default to using bf16 too, but modifying that would be at the risk of user and assuming they know what they are doing

sayakpaul:
Yeah makes sense. Do we wanna have a note about this in the docs?

sayakpaul:
I will open a PR.

PR #224, opened 2025-01-15
[docs] add a note on MP.

sayakpaul:
Fixes https://github.com/a-r-r-o-w/finetrainers/issues/222

### DIFF
--- a/docs/training/README.md
+++ b/docs/training/README.md
@@ -23,3 +23,12 @@ Supported training types:
 - Full finetuning (`--training_type full-finetune`)
 
 Arguments for training are well-documented in the code. For more information, please run `python train.py --help`.
+
+## How do we handle `mixed_precision`?
+
+The accelerate config files (the ones seen [here](../../accelerate_configs/)) that are being supplied while launching training should contain a field called `mixed_precision` and `accelerate` makes use of that if specified. We don't let users explicitly pass that from the CLI args because it can be confusing to have `transformer_dtype` and `mixed_precision` in the codebase.
+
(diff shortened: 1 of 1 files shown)


docs/training/README.md
```diff
@@ -23,3 +23,12 @@ Supported training types:
...
 
 Arguments for training are well-documented in the code. For more information, please run `python train.py --help`.
+
+## How do we handle `mixed_precision`?
+
+The accelerate config files (the ones seen [here](../../accelerate_configs/)) that are being supplied while launching training should contain a field called `mixed_precision` and `accelerate` makes use of that if specified. We don't let users explicitly pass that from the CLI args because it can be confusing to have `transformer_dtype` and `mixed_precision` in the codebase.
+
+`transformer_dtype` is the ultimate source of truth for the precision to be used when training. It will also most likely always have to be `torch.bfloat16` because:
+
+* All models currently supported (except Cog-2b) do not work well in FP16 for inference, so training would be broken as well. We can revisit if it makes sense to train in FP16 for other models added, but again, we don't really have to pass anything from our end since `accelerate` will read from config the file.
```
a-r-r-o-w, line 33:
```suggestion
* All models currently supported (except Cog-2b) do not work well in FP16 for inference, so training would be broken as well. This can be revisited if it makes sense to train in FP16 for other models added.
```

docs/training/README.md
```diff
@@ -23,3 +23,12 @@ Supported training types:
...
+* So, the `accelerate` config files now default to using "bf16" too, but modifying that would be at the risk of user and assuming they know what they are doing.
```
a-r-r-o-w, line 34:
```suggestion
* The `accelerate` config files default to using "bf16", but modifying that would be at the risk of user and assumes they understand the significance of their changes.
```

PR #225, opened 2025-01-16
[Chore] reset unnecessary args

sayakpaul:
When `training_type == "full-finetune"`, I think we should reset the default values of LoRA related args like `lora_alpha`, `rank`, etc. Otherwise, the logged args in the tracker might be confusing for the users. 

No need to explain if this is not right, I will close the PR. Thought of directly opening a PR as it was easier and not time-consuming at all.

Example overview: https://wandb.ai/sayakpaul/finetrainers-cog/runs/t1c5jcj3/overview.

### DIFF
--- a/finetrainers/args.py
+++ b/finetrainers/args.py
@@ -484,7 +484,7 @@ def validate_args(args: Args):
     _validated_model_args(args)
     _validate_training_args(args)
     _validate_validation_args(args)
-    _maybe_purge_unnecessary_args(args)
+    _maybe_reset_unnecessary_args(args)
 
 
 def _add_model_arguments(parser: argparse.ArgumentParser) -> None:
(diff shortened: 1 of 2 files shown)


a-r-r-o-w:
I think a better solution might be to only log relevant arguments based on training type, no? Would add a utility function like `get_training_info(args)` and log that

sayakpaul:
Will work on that and mark it ready for review when ready.

sayakpaul:
@a-r-r-o-w do the latest changes work for you?

Full-finetune logging: https://wandb.ai/sayakpaul/finetrainers-cog/runs/bzrif246/overview
LoRA logging: https://wandb.ai/sayakpaul/finetrainers-cog/runs/3m0vbhla/overview

a-r-r-o-w (approved):
Awesome, thanks!

PR #238, opened 2025-01-22
Condition

ArEnSc:


neph1:
Is this basically what is needed to train a control net for ltx? If so, are times comparable to lora training?

ArEnSc:
> Is this basically what is needed to train a control net for ltx? If so, are times comparable to lora training?

This is just me understanding what is going on with LTX not gonna lie I give up it doesn't work.

Issue #239, opened 2025-01-23
Change json dataset to a more manageable format


neph1:
### Feature request  / 功能建议

I'm struggling managing a dataset of only 10s of files, I can only imagine what it would be like with 100s of samples. Having path + prompt separate and unnumbered makes it difficult to ensure that each sample has the right prompt, or if one is missing (where?).
I propose changing the json dataset format to something like:

```
[
  {
    "type":"video",
    "path":"some/path",
    "prompt":"description"
  }
]
```

### Motivation / 动机

This makes it easy to also add additional information later on, if only for house-keeping, like "fps", and "frames", "width", etc etc.
Maybe even wrap the array in an additional layer, if a dataset could use other information too?

### Your contribution / 您的贡献

If the proposal is accepted, I can make a pr for this.

a-r-r-o-w:
Hey, JSON is already supported :) Well the code has a huge backlog pending for a refactor, but it works decently well for small datasets

https://github.com/a-r-r-o-w/finetrainers/blob/3eb9f8a6c4541e58ec5b3c7522ab003d2aa2d5f9/finetrainers/dataset.py#L195

You need to pass `--dataset_root` as the location where the videos are located, `--dataset_file` as the json file, `--video_column` as the name of the attribute that contains the path (path must be relative to dataset_root), and `--caption_column` as the attribute containing the prompt. LMK if something doesn't work and please feel free to modify the code and submit PRs to make it better suited for general use

PR #241, opened 2025-01-23
fix tests, small optimization to json dataset loading

neph1:
This fixes the dataset tests that were referencing the old cogvideox repo.
I also noticed while doing so that using ImageOrVideoDataset directly will throw an exception because max_num_frames is not set. So I moved that to __init__ of the base class.

There's also a minor refactor to the _load_dataset_from_json method, because I initially read it wrong and thought it was parsing two columns separately. At least it only parses the data once.

I still haven't done any test run with this. So maybe let it simmer for a bit.

Ref: #239

### DIFF
--- a/finetrainers/dataset.py
+++ b/finetrainers/dataset.py
@@ -246,7 +246,7 @@ class ImageOrVideoDataset(Dataset):
         video_reader = decord.VideoReader(uri=path.as_posix())
         video_num_frames = len(video_reader)
 
-        indices = list(range(0, video_num_frames))
+        indices = list(range(0, video_num_frames, video_num_frames // self.max_num_frames))
         frames = video_reader.get_batch(indices)
         frames = frames[: self.max_num_frames].float()
         frames = frames.permute(0, 3, 1, 2).contiguous()
--- a/finetrainers/dataset.py
+++ b/finetrainers/dataset.py
@@ -60,6 +60,8 @@ class ImageOrVideoDataset(Dataset):
         self.id_token = f"{id_token.strip()} " if id_token else ""
         self.resolution_buckets = resolution_buckets
 
+        self.max_num_frames = max(self.resolution_buckets, key=lambda x: x[0])[0]
+
         # Four methods of loading data are supported.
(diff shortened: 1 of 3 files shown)


finetrainers/dataset.py
```diff
@@ -139,7 +141,10 @@ def __getitem__(self, index: int) -> Dict[str, Any]:
         if video_path.suffix.lower() in [".png", ".jpg", ".jpeg"]:
             video = self._preprocess_image(video_path)
         else:
-            video = self._preprocess_video(video_path)
+            try:
+                video = self._preprocess_video(video_path)
+            except:
+                raise ValueError(f"Exception while preprocessing {video_path}.")
```
neph1, line 147: This was added for personal reasons, as I had difficulties finding out which file was causing issues

finetrainers/dataset.py
```diff
@@ -249,8 +258,6 @@ class ImageOrVideoDatasetWithResizing(ImageOrVideoDataset):
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
 
-        self.max_num_frames = max(self.resolution_buckets, key=lambda x: x[0])[0]
```
neph1, line 252: Moved to ImageOrVideoDataset

finetrainers/dataset.py
```diff
@@ -237,7 +246,7 @@ def _preprocess_video(self, path: Path) -> Tuple[torch.Tensor, Optional[torch.Te
         video_reader = decord.VideoReader(uri=path.as_posix())
         video_num_frames = len(video_reader)
 
-        indices = list(range(0, video_num_frames, video_num_frames // self.max_num_frames))
+        indices = list(range(0, video_num_frames))
```
neph1, line 249: I expected that in ImageOrVideoDataset, since it's not resizing, it shouldn't limit the number of frames, either? But in hindsight, maybe it should, since max_num_frames is set.

a-r-r-o-w:
Hi @neph1! Sorry about not reviewing this PR in time. I had a larger refactor in place so these changes would have been overwritten sooner or later. I hope you don't mind me closing the PR since the code is quite divergent now

PR #245, opened 2025-01-25
Model Specification API

a-r-r-o-w:
- Standardizes the implementation required for supporting new models. This makes it easier to implement arbitrary models - even ones that don't depend on Diffusers
- Introduces "Conditions". The goal here is to support many different opt-in configurations that allow users to use various default conditioning methods via CLI args. Good for quick prototyping new ideas with models and helpful for upcoming Control LoRA support
- Allows usage of arbitrary text encoders, tokenizers, transformers outside of the `pretrained_model_name_or_path`
- Allows more flexibility in the calculation of prediction and targets, so that we don't have to workaround for models like Mochi

The goal here is to abstract out parts of the the trainer that _could possibly be_ model-specific. The overall training algorithm, distributed-ness, etc. will always still be handled by the trainer. There are still some kinks to work out but it is getting there slowly.

The next step, either in this PR or follow-up, or in near future depending on if I find the time this weekend, is to remove `accelerate` usage. After several hours of debugging and trying to make context parallism work with it, I can't get around NCCL hangs. Pure pytorch is just plain simpler to work with, and the way for using all the goodies of parallism-types to make training more scalable. Things may possibly break or be unstable for a bit, so planning on doing a v0.0.1 release before I merge this PR.

TODO:
- [ ] Convert CogVideoX
- [ ] Convert HunyuanVideo

### DIFF
--- a/finetrainers/args.py
+++ b/finetrainers/args.py
@@ -4,6 +4,7 @@ from typing import Any, Dict, List, Optional, Tuple
 
 import torch
 
+from .conditions import SUPPORTED_CONDITIONS, ConditionType
 from .constants import DEFAULT_IMAGE_RESOLUTION_BUCKETS, DEFAULT_VIDEO_RESOLUTION_BUCKETS
 from .models import SUPPORTED_MODEL_CONFIGS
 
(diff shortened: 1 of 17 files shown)


a-r-r-o-w:
Actually, no urgent reason to merge this soon. Will get DDP/FSDP, PP and TP working together atleast, since if we are going to be flaky for a bit, it's best to have all features available for testing and work my way through bugs

PR #251, opened 2025-01-29
fix: model card info.

sayakpaul:
* Save model cards when requested even when there's no validation prompt.
* Logging of the ID_token in the model card.
* Better tags for model card.

Example: https://huggingface.co/sayakpaul/dummy-ltx

### DIFF
--- a/finetrainers/utils/hub_utils.py
+++ b/finetrainers/utils/hub_utils.py
@@ -28,16 +28,19 @@ def save_model_card(
                 }
             )
 
+    training_type = "Full" if args.training_type == "full-finetune" else "LoRA"
     model_description = f"""
-# LoRA Finetune
+# {training_type} Finetune
 
 <Gallery />
 
 ## Model description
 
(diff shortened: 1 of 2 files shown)


finetrainers/utils/hub_utils.py
```diff
@@ -69,9 +72,13 @@ def save_model_card(
         "text-to-video",
         "diffusers-training",
         "diffusers",
-        "lora",
+        "finetrainers",
         "template:sd-lora",
     ]
+    if training_type == "Full":
```
a-r-r-o-w, line 78: This should be `full-finetune`, no?

sayakpaul: We are setting the `training_type` to include in the model card here:
https://github.com/a-r-r-o-w/finetrainers/pull/251/files#diff-2cde8b28954d58818f7f63885f42609611c8285e38c6b6956d49d910869f6465R31

```
training_type = "Full" if args.training_type == "full-finetune" else "LoRA"
```

a-r-r-o-w (approved):

PR #253, opened 2025-01-30
[chore] relax requirements a bit.

sayakpaul:
Until and unless a user needs it, I think it is okay to not install `torchao` and `bitsandbytes` by default.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,4 @@
 accelerate
-bitsandbytes
 diffusers>=0.32.1
 transformers>=4.45.2
 huggingface_hub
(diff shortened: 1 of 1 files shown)


a-r-r-o-w:
Will keep this open for a bit if you don't mind. In #245, I'm also exploring using torchao fp8 directly since that would be true fp8 training instead of what we have at the moment, so maybe not too problematic to have it as a direct dependancy. Our layerwise implementation is good for most GPUs but, in the end, it is unstable. TorchAO fp8 would be true fp8 and more stable, but currently limited to Ada and Hopper only I think

PR #275, opened 2025-03-02
calling save_model_card correctly

vivkul:
Previously, when `num_validation_samples == 0` then `save_model_card` was called even if `self.args.push_to_hub` or `final_validation` was `False`.


a-r-r-o-w (approved):
Thanks for the fix!

PR #290, opened 2025-03-04
Add 3DGS dataset example for Wan

a-r-r-o-w:
From a small set of experiments (~10), I've found it hard to train in new concepts to the 1.3B model. It ultimately leads to model collapsing and performing bad with the lora. This is my best results so far:

https://github.com/user-attachments/assets/66d74fcb-999d-4014-847a-ef7abd74f0c7

Still not going for training the 14B variants because we need to make training significantly faster. The following are on my mind to support quickly:
- PTD experimental context parallel
- Flash Attention natively in diffusers (WIP)

### DIFF
--- a/examples/training/sft/wan/3dgs_dissolve/train.sh
+++ b/examples/training/sft/wan/3dgs_dissolve/train.sh
@@ -1,0 +1,162 @@
+#!/bin/bash
+
+set -e -x
+
+# export TORCH_LOGS="+dynamo,recompiles,graph_breaks"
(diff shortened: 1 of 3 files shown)


a-r-r-o-w:
Checkpoint out: https://huggingface.co/finetrainers/Wan2.1-T2V-1.3B-3dgs-v0

Some more samples:

https://github.com/user-attachments/assets/19705b7a-b7ea-46ab-91d5-09630e330ab9

PR #310, opened 2025-03-10
Channel-concatenated Control Trainer

a-r-r-o-w:
This PR introduces a channelwise-concat control conditioning trainer. It involves adding extra input channels to the patch embedding layer (referred to as the "control injection" layer in finetrainers), to mix conditioning features into the latent stream. This architecture choice is very common and has been seen before in many models - CogVideoX-I2V, HunyuanVideo-I2V, Alibaba's Fun Control models, etc. Due to the popularity and simplicity in the architecture choice, it is a good choice to support standalone as a trainer.

The PR introduces Control condition training for CogView4 and Wan. More models will be supported in future.

Typically, inference would require writing a full pipeline and integration with diffusers for ease of use. However, one goal with finetrainers is to minimize the burden for upstream libraries and downstream applications. That is, you should be able to run a finetrainers trained model without it actually having to be supported anywhere else. So, some constructs are minimally implemented until libraries/applications decide to add native support. For example, here's how you can run CogView4 inference without it having upstream support in diffusers:

```python
import torch
from diffusers import CogView4Pipeline
from diffusers.utils import load_image
from finetrainers.models.utils import _expand_linear_with_zeroed_weights
from finetrainers.patches import load_lora_weights
from finetrainers.patches.dependencies.diffusers.control import control_channel_concat

dtype = torch.bfloat16
device = torch.device("cuda")
generator = torch.Generator().manual_seed(0)

pipe = CogView4Pipeline.from_pretrained("THUDM/CogView4-6B", torch_dtype=dtype)

in_channels = pipe.transformer.config.in_channels
patch_channels = pipe.transformer.patch_embed.proj.in_features
pipe.transformer.patch_embed.proj = _expand_linear_with_zeroed_weights(pipe.transformer.patch_embed.proj, new_in_features=2 * patch_channels)

load_lora_weights(pipe, "finetrainers/CogView4-6B-Edit-LoRA-v0", "cogview4-lora")
pipe.set_adapters("cogview4-lora", 0.9)
pipe.to(device)

prompt = "Make the image look like it's from an ancient Egyptian mural."
control_image = load_image("examples/training/control/cogview4/omni_edit/validation_dataset/0.png")
height, width = 1024, 1024

with torch.no_grad():
    latents = pipe.prepare_latents(1, in_channels, height, width, dtype, device, generator)
    control_image = pipe.image_processor.preprocess(control_image, height=height, width=width)
    control_image = control_image.to(device=device, dtype=dtype)
    control_latents = pipe.vae.encode(control_image).latent_dist.sample(generator=generator)
    control_latents = (control_latents - pipe.vae.config.shift_factor) * pipe.vae.config.scaling_factor

with control_channel_concat(pipe.transformer, ["hidden_states"], [control_latents], dims=[1]):
    image = pipe(prompt, latents=latents, num_inference_steps=30, generator=generator).images[0]

image.save("output.png")
```

Obviously, this is not a great solution but it's better to have something be runnable than not being able to run at all unless native support is added elsewhere, in my personal opinion.


a-r-r-o-w:
Yes, exactly that. Loras (and more eventually, like Lycoris, Lokr, other peft methods) and full-finetune support.

The PR is ready to merge except for one silly bug which makes training validation look good, but any attempt to use the lora after training generates random noise. I'm trying to figure it out with the help of a teammate so we can expect this to be merged very soon.

neph1:
As an avid controlnet user, I'm really looking forward to this feature.

a-r-r-o-w:
After a lot of banging-head-on-desk, the bug isn't really in finetrainers and rather comes from how lora weights config is autodetermined and used in peft. I'll open upstream PRs but not too happy about how long it took for a simple trainer addition.

Anyway, this was good learnings and something I can keep in mind for future improvements and newer training algorithms that support low rank techniques

a-r-r-o-w:
### Canny

Easier to train. Signs of fast converging

<table align=center>
<tr>
  <td align=center></td>
</tr>
<tr>
  <td align=center></td>
</tr>
<tr>
  <td align=center></td>
</tr>
<tr>
  <td align=center></td>
</tr>
</table>

These results come from only 2500 training steps. My run died due to an unrelated memory issue on shared GPU node. The original goal was to train for 10000 steps but I think I might have to postpone it for a while since there are more important goals with the project to focus on.

The loss looks stable and grad_norm lowers over time hinting towards convergence.

image

image

### Editing

Harder to train and seems to be taking long to converge. Can do bigger runs once we're faster and more optimized. It seems attention masks for text embeddings may play an important role. The recommendation from the team is to train with high batch size for 30k+ steps for good convergence

CogView4 Control Edit v0 checkpoint: https://huggingface.co/finetrainers/CogView4-6B-Edit-LoRA-v0

PR #318, opened 2025-03-11
add valid names to dataset docs

neph1:
I figured this might save people a few minutes of troubleshooting.

I would suggest, though, to allow any name through the dataset config json


a-r-r-o-w (approved):
Thanks!

PR #319, opened 2025-03-12
Add support for hunyuan-i2v

tolgacangoz:
I am quite new in video models and just started to read the `finetrainers`' codebase yesterday. So, this PR is open to all types of collabs 🤝

- https://github.com/huggingface/diffusers/pull/10983
- https://github.com/Tencent/HunyuanVideo-I2V#-news
- https://github.com/Tencent/HunyuanVideo-I2V/blob/main/train_image2video_lora.py

### DIFF
--- a/finetrainers/models/hunyuan_video/base_specification.py
+++ b/finetrainers/models/hunyuan_video/base_specification.py
@@ -7,9 +7,11 @@ from diffusers import (
     AutoencoderKLHunyuanVideo,
     FlowMatchEulerDiscreteScheduler,
     HunyuanVideoPipeline,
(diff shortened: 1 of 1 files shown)


a-r-r-o-w (commented):
Thanks @tolgacangoz! I believe the PR currently enables validation when an image is provided. However, it still will not work due to the transformer model for I2V task being different from T2V task.

Training support will require writing a new full model specification I believe (not sure if you had that planned and this is too early for a review 😅, so sorry if I'm being stupid)

tolgacangoz:
I decided to return shortly after studying video models more (plus the GPU mode course).

Issue #295, opened 2025-03-05
Memory using Validation


dorpxam:
I've just update with the latest version and start a training (LTXV LoRA) using validation steps this time. I use the new bash method (from the `sft/ltx_video/crush_smol_lora` example). I've some details to report. 

1) The truncation of the prompt used for validation with the message : `The following part of your input was truncated because 'max_sequence_length' is set to  128 tokens`. Is the max_sequence_length customizable ?
2) The generated validation video are in the filename format : `validation-step-n-n-idtoken-startoftheprompt.mp4`. The problem is that if two validation prompts start with the same sequence, one validation video overwrite the other. Maybe it can be cool to replace the 'startoftheprompt' text by the id of the prompt in the list or a hash ?
3) Here with my RTX4080 (16GB), I do not have OOM using the validation steps, but the problem comes with the release of the memory after the validation. Here is the logs :

```
INFO - Memory before training start: {
    "memory_allocated": 3.911,
    "memory_reserved": 4.463,
    "max_memory_allocated": 4.021,
    "max_memory_reserved": 4.463
}

INFO:finetrainers:Memory before training start: {
    "memory_allocated": 3.911,
    "memory_reserved": 4.463,
    "max_memory_allocated": 4.021,
    "max_memory_reserved": 4.463
}

INFO - Training configuration: {
    "trainable parameters": 29360128,
    "train steps": 1000,
    "per-replica batch size": 1,
    "global batch size": 1,
    "gradient accumulation steps": 1
}

INFO - Memory before validation start: {
    "memory_allocated": 3.935,
    "memory_reserved": 6.84,
    "max_memory_allocated": 12.823,
    "max_memory_reserved": 24.469
}

INFO:finetrainers:Memory before validation start: {
    "memory_allocated": 3.935,
    "memory_reserved": 6.84,
    "max_memory_allocated": 12.823,
    "max_memory_reserved": 24.469
}

INFO - Memory after validation end: {
    "memory_allocated": 13.614,
    "memory_reserved": 14.135,
    "max_memory_allocated": 15.019,
    "max_memory_reserved": 24.469
}

INFO:finetrainers:Memory after validation end: {
    "memory_allocated": 13.614,
    "memory_reserved": 14.135,
    "max_memory_allocated": 15.019,
    "max_memory_reserved": 24.469
}
```
[...]

PR #320, opened 2025-03-13
Fix enable_model_cpu_offload problems

### LINKED ISSUE #212: ``enable_model_cpu_offload`` causes NCCL timeout during multi-gpu training
Hi, finetrainers team.

I ran ``finetrainers/training/train_text_to_video_sft.sh`` with ``--enable_model_cpu_offload`` on 4 A100 80GB gpus (single node). However, after validation, all the processes were blocked instead of running the next training step. Without the argument, everything works fine.

a-r-r-o-w:
Fixes #212, #295

cc @dorpxam Could you give this a try? I believe it should fix any issues you're facing when using enable_model_cpu_offload

### DIFF
--- a/finetrainers/args.py
+++ b/finetrainers/args.py
@@ -853,8 +853,9 @@ def _validate_dataset_args(args: BaseArgs):
 
 
 def _validate_validation_args(args: BaseArgs):
-    if args.world_size > 1 and args.enable_model_cpu_offload:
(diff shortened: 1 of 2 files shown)


dorpxam:
Just see that one. For me precomputation is not more an issue with latest updates. Thanks so much!

PR #328, opened 2025-03-15
Json dataset loading and testing

neph1:
As it stands, tests will fail since there are two metadata files in the folder (one csv as well)


a-r-r-o-w (commented):
Hmm... not quite sure why this doesn't already load. ImageFolderDataset and VideoFolderDataset should've autodetected this.

Could be a bug, I'll take a look asap

PR #351, opened 2025-03-25
add 🧨diffusers inference example to model card

linoytsaban:
Hey! I think it can be great to have the model cards automatically uploaded with the relevant inference example. 
WDYT about something like this?

### DIFF
--- a/docs/models/hunyuan_video.md
+++ b/docs/models/hunyuan_video.md
@@ -22,9 +22,6 @@ Assuming your LoRA is saved and pushed to the HF Hub, and named `my-awesome-name
 
 ```py
 import torch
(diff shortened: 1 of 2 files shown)


a-r-r-o-w:
Sorry for the delay! I missed all notifications for PRs that happened during diffusers offsite haha, taking a look

a-r-r-o-w (approved):
Changes LGTM! We might have to refactor a bit in the future though, to support multiple upcoming trainers I've been working on (channel-concatenation, VideoJAM, etc.). Would love your thoughts on the same when we get to it :)

a-r-r-o-w:
@linoytsaban Could you run `make style` 😅

PR #354, opened 2025-03-30
Control lora trainer hunyuan

neph1:
i tried to follow how the wan control specification looked and apply it to hunyuan video. ~~it runs, but i OOM on 24GB, even though i apply "layerwise upcasting". maybe i'm doing it wrong, or does it require more vram than normal training?~~
Yeah, so I haven't been able to verify the results.
There was a lot of copied code, so I moved what I could to a common file instead.
This is not the training script i used, I tried to make it similar to the other tranining scripts before pushing.


finetrainers/models/hunyuan_video/control_specification.py
```diff
@@ -0,0 +1,358 @@
...
+            )
+        else:
+            transformer = HunyuanVideoTransformer3DModel.from_pretrained(
+                self.pretrained_model_name_or_path,
+                subfolder="transformer",
+                torch_dtype=self.transformer_dtype,
+                **common_kwargs,
+            )
+
+        # TODO(neph1): Need to update the patch embed layer
```
neph1, line 108: I've made some progress but could use some advice here. I believe I need to update this, but I'm not sure how. The line itself is taken from HunyuanVideoTransformer3DModel.

Or maybe I'm completely off track.

finetrainers/models/hunyuan_video/control_specification.py
```diff
@@ -0,0 +1,358 @@
...
+        # transformer.x_embedder = HunyuanVideoPatchEmbed((patch_size_t, patch_size, patch_size), new_in_features)
```
neph1, line 109: I've made some progress but could use some advice here. I believe I need to update this, but I'm not sure how. The line itself is taken from HunyuanVideoTransformer3DModel.

Or maybe I'm completely off track.

neph1:
Omg, it actually runs.

PR #360, opened 2025-04-07
fix string matching for blocks

neph1:
This is not complete, but I need some help, here.
It fixes the string matching for the blocks.
But the 'proj_mlp' key is nowhere to be found in the loras. I've tried various settings, and observed that the size of the lora changes, but neither has a key like that.
Lora config seems to be set properly, and applied.
As far as I can see, it's up to diffusers to save everything, so maybe it's something there.

Removing the proj_mlp from the concat makes the lora work, but only the double blocks. The single blocks have a mismatching size.

### DIFF
--- a/examples/formats/hunyuan_video/convert_to_original_format.py
+++ b/examples/formats/hunyuan_video/convert_to_original_format.py
@@ -17,10 +17,13 @@ def convert_lora_sd(diffusers_lora_sd):
 
     prefix = "diffusion_model."
 
+    double_block_pattern = "transformer.transformer_blocks"
+    single_block_pattern = "transformer.single_transformer_blocks"
+
     converted_lora_sd = {}
     for key in diffusers_lora_sd.keys():
         # double_blocks
-        if key.startswith("transformer_blocks"):
+        if key.startswith(double_block_pattern):
             # img_attn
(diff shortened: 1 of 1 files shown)


a-r-r-o-w:
Hmm, so I tried porting over their original code and I'm not quite happy about the amount of additional code that needs to be added to the repository for supporting it natively (and ofcourse the additional maintainence burden). It's about 2000 LoC for the modeling and another 500 for the ModelSpec + ~300 for additional config management and weight loading stuff. I don't really want to use git submodules in the project either.

Perhaps, it's just easier to try and make the conversion script work as best as possible

neph1:
tbh, I'm questioning whether the conversion script should live in this repo, at all.
this project provides functionality to train models, and produces a common format for its output. how that is used in different inference engines could be handled outside.
in my use case, i build loras. i have a way of using them, and i would like others to be able to use them, too. hence my wish for the script to make them work in comfy ui. but i don't think you should spend time on that. maybe it can be lifted to an external repo.
with the information you've given, i'll probably spend some more time on fixing it up, if not only to see how lossy the conversion is (chances are my on workflow is just as lossy).

neph1:
I now get equal results to my own ComfyUI fork. So I would consider this done. Size of the lora is *4, though.

PR #364, opened 2025-04-10
[Fix] Raise ValueError proactively before some confusing errors occur due to wrong input image size

DarkSharpness:
## Motivation

When I was finetuning a flux-dev model with my custom dataset, I got the following error:

```
Calculated padded input size per channel: (2 x 321). Kernel size: (3 x 3). Kernel size can't be greater than the actual input size
```

It was confusing, and it turned out that the image in my dataset was smaller than the crop size, which eventually led to the kernel error.

## Solution

I suggest we raise an exception when the input image/video is of the wrong size.

Moreover, I think we may add an option to keep only those images whose sizes are no smaller than the crop size and drop the others. (if needed, I can do that)

### DIFF
--- a/finetrainers/functional/image.py
+++ b/finetrainers/functional/image.py
@@ -8,9 +8,7 @@ def center_crop_image(image: torch.Tensor, size: Tuple[int, int]) -> torch.Tenso
     num_channels, height, width = image.shape
     crop_h, crop_w = size
     if height < crop_h or width < crop_w:
-        raise ValueError(
-            f"Image size {(height, width)} is smaller than the target size {size}."
-        )
+        raise ValueError(f"Image size {(height, width)} is smaller than the target size {size}.")
     top = (height - crop_h) // 2
     left = (width - crop_w) // 2
     return image[:, top : top + crop_h, left : left + crop_w]
--- a/finetrainers/functional/image.py
+++ b/finetrainers/functional/image.py
@@ -7,6 +7,10 @@ import torch.nn.functional as F
 def center_crop_image(image: torch.Tensor, size: Tuple[int, int]) -> torch.Tensor:
     num_channels, height, width = image.shape
     crop_h, crop_w = size
+    if height < crop_h or width < crop_w:
+        raise ValueError(
+            f"Image size {(height, width)} is smaller than the target size {size}."
+        )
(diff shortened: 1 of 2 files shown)


a-r-r-o-w:
@DarkSharpness Could you give me an example of what `--image_resolution_buckets` and image size would cause it to fail? Would like to understand/replicate the issue better before merging

DarkSharpness:
Sure. A similar error can be reproduced by modifying `examples/training/sft/flux_dev/raider_white_tarot/training.json` as follows:

```json
{
  "datasets": [
    {
      "data_root": "./boat",
      "dataset_type": "image",
      "id_token": "test",
      "image_resolution_buckets": [
        [1280, 1280]
      ],
      "reshape_mode": "center_crop",
      "remove_common_llm_caption_prefixes": true
    }
  ]
}
```

Then, running the script `examples/training/sft/flux_dev/raider_white_tarot/training.sh` on my H100 machine results in the following error:

```
RuntimeError: Calculated padded input size per channel: (2 x 321). Kernel size: (3 x 3). Kernel size can't be greater than actual input size
```

This error is caused by a missing boundary check when performing center cropping. In this specific case, the bucket `[1280, 1280]` is invalid, since the input images are sized `(1144 x 1919)`—the width is smaller than the target crop width. When `height < crop_h` or `width < crop_w`, the `top` and `left` values in the following cropping function can go negative, leading to unexpected behavior or runtime errors:

```python
def center_crop_image(image: torch.Tensor, size: Tuple[int, int]) -> torch.Tensor:
    num_channels, height, width = image.shape
    crop_h, crop_w = size
    top = (height - crop_h) // 2
    left = (width - crop_w) // 2
    return image[:, top : top + crop_h, left : left + crop_w]
```

While it's reasonable to assume users should provide bucket sizes that do not exceed the input image dimensions, I believe we should still proactively detect and reject incompatible image sizes early to avoid cryptic runtime errors.

a-r-r-o-w (approved):
Thanks for the explanation, it makes sense to me now! I think we should document this in the dataset docs (only for `center_crop` mode, I believe) because otherwise people may experience failure in between training and lose essential progress. Would you like to open a PR making a note of the same?

PR #369, opened 2025-04-12
Feature/hunyuan control lora v2

neph1:
I believe this is live again. It does stuff.
I will try to find a small (canny?) dataset and run a limited test run to see if it converges.
I have no idea how to inference, though. Any example scripts I can modify?


finetrainers/models/hunyuan_video/base_specification.py
```diff
@@ -57,19 +57,25 @@ def forward(
         assert video.ndim == 5, f"Expected 5D tensor, got {video.ndim}D tensor"
         video = video.to(device=device, dtype=vae.dtype)
         video = video.permute(0, 2, 1, 3, 4).contiguous()  # [B, F, C, H, W] -> [B, C, F, H, W]
-
+        
+        compute_posterior = False
```
neph1, line 61: Only works with compute_posterior = False, for now

PR #338, opened 2025-03-19
Wan Control Trainer

a-r-r-o-w:
This PR adds support for channel-concatenated control condition training for Wan, and it is part of #310.

With video models, you get another dimension of controllability. Recently, we've seen a variety of applications that support interpolating between images, adding effects, advanced controllability by extra input signals, video extension, image conditioning, etc. The goal with this PR is to enable all of that with this simple architecture modification. See #310 for more details.

This PR introduces `FrameConditioningType`, which is specifically used only for video models. The following conditioning types are currently supported:
- INDEX: Condition on any arbitrary latent frame. The only sensible choice is to set the `--frame_conditioning_index` to `0`, but it is simple to implement for any frame and enables your wacky lora ideas to come true!
- PREFIX: Condition on some number of start frames. This is most similar to NVIDIA's Cosmos Video2World model, where you can extend an image/video
- RANDOM: Condition on randomly selected frames. This technique allows multiple images to be used for interpolation. Give your crazy lora ideas a life!
- FIRST_AND_LAST: Condition only on first and last frames for interpolating between two images.
- FULL: This is the default setting. It's best used when you want to condition on an entire different video for generation. For example, providing an input video for editing, or canny/depth/pose control.

Additionally, the option `--frame_conditioning_concatenate_mask` can be used to add a single channel frame mask indicating to the model what frames are used for conditioning. It has not been rigourously tested, but may help with model convergence. More experiments need to be carried out to understand its effectiveness.

### Inference

The following results are from a 10k step run on a single GPU with first frame conditioning. The other runs with different conditioning types also succeeded and seem to be working well. However, 10k steps with batch_size=1 is too low for convergence and the model does not generalize well at this scale. Bigger runs are needed for releasable checkpoints, and will be queued after the trainer is made faster!

```python
import numpy as np
import torch
from diffusers import WanPipeline
from diffusers.utils import export_to_video, load_video
from finetrainers.trainer.control_trainer.data import apply_frame_conditioning_on_latents
from finetrainers.models.utils import _expand_conv3d_with_zeroed_weights
from finetrainers.patches import load_lora_weights
from finetrainers.patches.dependencies.diffusers.control import control_channel_concat

dtype = torch.bfloat16
device = torch.device("cuda")
generator = torch.Generator().manual_seed(0)

pipe = WanPipeline.from_pretrained("Wan-AI/Wan2.1-T2V-1.3B-Diffusers", torch_dtype=dtype).to(device)

in_channels = pipe.transformer.config.in_channels
[...]

PR #373, opened 2025-04-12
Control lora trainer hunyuan

neph1:
Still have some cleaning up to do, but this pr is back (I might migrate this and force push, or if it's eventually squashed).
It does things and produced something. I will do a longer run tomorrow, but I have no idea how to inference. ~Is there an example script somewhere i can modify?~ Found it


finetrainers/models/hunyuan_video/base_specification.py
```diff
@@ -57,19 +59,25 @@ def forward(
         assert video.ndim == 5, f"Expected 5D tensor, got {video.ndim}D tensor"
         video = video.to(device=device, dtype=vae.dtype)
         video = video.permute(0, 2, 1, 3, 4).contiguous()  # [B, F, C, H, W] -> [B, C, F, H, W]
-
+        
+        compute_posterior = False
```
neph1, line 63: So far only made it work with compute_posterior false

neph1:
I will try a first frame conditioning run, since I believe using the hunyuan i2v pipeline would be able to verify if it works or not. My current test run was for canny using hunyuan t2v, which was probably a mistake.

a-r-r-o-w (commented):
Hey @neph1, thanks for the PR and testing with some runs! I know there's some duplication of code at the moment, but I plan to address that in the future with something else. For now, let's keep the duplication

I'll try to help with launching some training runs too for verifying the PR once we have the changes looking similar to the Wan/CogView control implementations #338

neph1:
Oh, I did follow the new implementation, just kept it on the same branch. At least I believe it follows what is in main now.
Not sure it works, though. I ran 1000 steps canny training on some simpsons dataset I found, think I managed to make the conversion script handle x_embedder, but it's only noise. Sadly can't inference with plain diffusers due to VRAM.

Got a bit disheartened, but I'll get back to it.

finetrainers/models/hunyuan_video/base_specification.py
```diff
@@ -58,18 +58,24 @@ def forward(
...
+            # if vae.use_slicing and video.shape[0] > 1:
+            #     encoded_slices = [vae._encode(x_slice) for x_slice in video.split(1)]
+            #     moments = torch.cat(encoded_slices)
+            # else:
+            #     moments = vae._encode(video)
+            moments = vae._encode(video)
             latents = moments.to(dtype=dtype)
 
-        return {self.output_names[0]: latents}
+        latents_mean = torch.tensor(vae.latent_channels)
```
a-r-r-o-w, line 75: @neph1 These changes seem incorrec to me and will cause worse generations. The previous implementation that did not perform this normalization was correct, I think.

Was this modified from Wan? If so, it's incorrect because they are different models and preprocess latents differently

neph1:
By all means, if you have the time. :+1:

Issue #267, opened 2025-02-19
How to save the best performing checkpoint during LoRA fine-tuning on Hunyuan Video?


dingangui:
In the HunyuanVideo training scripts, we can save checkpoints every 500 steps by passing `--checkpointing_steps 500`. The final model is saved through the following code:

```python
if accelerator.is_main_process:
    transformer = unwrap_model(accelerator, self.transformer)

    if self.args.training_type == "lora":
        transformer_lora_layers = get_peft_model_state_dict(transformer)

        self.model_config["pipeline_cls"].save_lora_weights(
            save_directory=self.args.output_dir,
            transformer_lora_layers=transformer_lora_layers,
        )
    else:
        transformer.save_pretrained(os.path.join(self.args.output_dir, "transformer"))
```
(Reference: https://github.com/a-r-r-o-w/finetrainers/blob/4bb10c62324aef4fbac85bb381acb9f6f39a5076/finetrainers/trainer.py#L837C1-L848C95)

My question is: How can I ensure that I save the best performing model during LoRA fine-tuning? The final saved model might not be the best, as the loss could fluctuate during training. The same applies to intermediate checkpoints. Is there a recommended approach for tracking and saving the best-performing model?

neph1:
You test them yourself and see. Or you use the validation videos if you're fortunate enough to be able to generate them. You look for the best convergence with your concept on a range of subjects.
I haven't been able to overfit so far, but I'm only on a measly 3090.

dingangui:
> You test them yourself and see. Or you use the validation videos if you're fortunate enough to be able to generate them. You look for the best convergence with your concept on a range of subjects. I haven't been able to overfit so far, but I'm only on a measly 3090.

Thanks for your reply, I'm doing exactly what you said, I checked the validation videos and found that results from the later checkpoints sometimes perform worse than the earlier ones. So I’m worried that the best model might not be the latest one.

BlackTea-c:
[...]

PR #379, opened 2025-04-22
chore: save all weights with step-specific directories

Leojc:
Maybe we should keep all the weights? #267


a-r-r-o-w (approved):
Thanks, makes sense!

The checkpointer implementation also saves the entire training state into the output directory, but as a single file. I will try to see if can instead split the optimizer and other states from the state dict and store it individually. Otherwise, the disk usage will increase rapidly due to strong two copies of the same thing in different format

PR #382, opened 2025-04-23
fix: lora loading for final validation

Leojc:
Forgot to change lora loading accordingly. #379


a-r-r-o-w (approved):
Thanks for the fix! I forgot to run the tests before merging 😅

PR #396, opened 2025-05-03
docker file add

ootsuka-repos:
windows用にdocker fileを追加しdockerで動作するようにしました。


a-r-r-o-w:
Thanks for the PR @ootsuka-repos! Could we move this content related to Docker into `examples/docker` folder please? Would like to keep the root directory very minimal. Also, it would be preferable to have instructions in english since it is common language for most users, but we can keep a chinese README as well

Gentle ping @zRzRzRzRzRzRzR if you'd like to give a review and help with translating chinese instructions to english

ootsuka-repos:
@a-r-r-o-w 
The file has been moved.
Also, comments were changed to English.

PR #397, opened 2025-05-06
Fix Wan2.1 I2V fine-tuning

detkov:
I tried to fine-tune Wan2.1 I2V as it is shown in the `examples/training/sft/wan_i2v/3dgs_dissolve/train.sh`. I downloaded datasets, etc, it was okay.

The issue I've faced was related to the [finetrainers/models/wan/base_specification.py](https://github.com/a-r-r-o-w/finetrainers/blob/main/finetrainers/models/wan/base_specification.py#L240).
It basically said that `self.transformer_config.pos_embed_seq_len` doesn't exist.
When I changed this to `self.transformer_config.get('pos_embed_seq_len')`, the issue was resolved.

FYI:
diffusers version is 0.34.dev (86294d3c7ffa5e97b9483d82c14021d06b701a12)
torch1.7.0+cu128
NVIDIA H100 PCIe

This PR addresses this problem and solves it.

### DIFF
--- a/finetrainers/models/wan/base_specification.py
+++ b/finetrainers/models/wan/base_specification.py
@@ -237,7 +237,7 @@ class WanModelSpecification(ModelSpecification):
             cache_dir=cache_dir,
         )
 
-        use_last_frame = self.transformer_config.pos_embed_seq_len is not None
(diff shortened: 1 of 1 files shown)


a-r-r-o-w (approved):
Thanks, makes sense!

Issue #412, opened 2025-06-04
ModuleNotFoundError: No module named 'finetrainers.models._metadata'


jbilcke-hf:
### System Info / 系統信息

When installing Finetrainers as a python module using PIP, the content of the `_metadata` is missing:

```
ModuleNotFoundError: No module named 'finetrainers.models._metadata'
```

Solution: this appears to be caused by the lack of a `__init__.py` in the `_metadata/` folder, making it being ignored by PIP during installation

### Reproduction / 复现过程

this can be reproduced by installing the latest version of Finetrainers from the git repo:
```
finetrainers @ git+https://github.com/a-r-r-o-w/finetrainers.git@main
```

### Expected behavior / 期待表现

We should be able to import 'finetrainers.models._metadata' after a PIP install

PR #413, opened 2025-06-04
Create __init__.py in finetrainers/models/_metadata/

jbilcke-hf:
This should fix the module not found error in #412


a-r-r-o-w (approved):
Awesome, thanks!

PR #414, opened 2025-06-04
add missing __init__.py files

jbilcke-hf:
Sorry I didn't notice in my previous PR that there were also new modules in the `dependencies/*`

I've created them as empty files, this should be enough for PIP


a-r-r-o-w (approved):
Nice, thanks!

PR #425, opened 2025-09-18
Torch 2.8 and local dataset

paul-at-nangalan:
Added some changes that seem to be needed for local file pair data set and for torch 2.8 (VideoReader seems to have been depricated)


paul-at-nangalan:
didn't mean to merge to HF repo

